PR #12, opened 2015-01-13
[#2629] Compilator choice in AIX and Solaris.

dumol:
Problem?
------------
Our Python package is considerably slower in AIX compared to other OS'es. Reading up on problem performances in AIX for other free software products suggests compiling with GCC might considerably improve the situation.

Solution?
-------------
Make it possible to compile `python-package` with GCC in AIX and Solaris so that we can test this hypothesis.

**Drive-by fix**:
  * consolidated `ARCH` renaming in `paver.sh`.

How to test?
-----------------
Re-compile `python-package` with GCC in AIX/Solaris by commenting the lines exporting `CC` and `CXX`. 
Re-run server tests with the new package in `upload/testing`.

**Notes**: 
  * on AIX this requires a newer GCC than the one available from IBM 
  * only the AIX 5.3 vWPAR was tested with GCC, third-party RPMs were used
  * only the included GCC3 was used to test Solaris 10 and only for a 64bit build
  * only the package compiled with GCC on AIX 5.3 was tested with server tests and it turned out to be about the same speed.

reviewer: @adiroiban

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -40,15 +40,25 @@ fi
 PYTHON_VERSION=`cut -d' ' -f 2 DEFAULT_VALUES`
 OS=`cut -d' ' -f 3 DEFAULT_VALUES`
 ARCH=`cut -d' ' -f 4 DEFAULT_VALUES`
+CC=`cut -d' ' -f 5 DEFAULT_VALUES`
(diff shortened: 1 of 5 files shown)


dumol:
**needs-review**

adiroiban:
Looks good. Thanks!

approved-at 4dfb37d

PR #13, opened 2015-01-22
[#2647] Generic Linux x86 support.

dumol:
Problems?
------------
We don't have an **x86** slave for `server`, `python-package` etc.
We equate unsupported/untested Linux distros with old Ubuntu LTS distros.

Solution?
-------------
Added a Debian 7 **x86** slave for `brink`, `python-package`, `compat`, `empirical`, `watchdog` and `server` and updated `python-package` with generic Linux support

**Drive-by fixes**:
  * added automatic removal of packages installed as dependencies during build in Ubuntu, RHEL and recent SLES distros
  * reshaped OS detection/naming to differentiate among old releases of supported OS'es and unsupported OS'es. Newer releases of supported OS'es shouldn't require any special addition in `paver.sh` from now on
  * removed `libbz2` headers installation
  * clean-up the install sub-directory to avoid contamination of new builds
  * check for unwanted dependencies in Linux, AIX, Solaris and OS X
  * edited `setup.py` to prevent building the following modules: `bz2`, `_tkinter`, `_curses` , `_curses_panel`, `_sqlite3`.

How to test?
----------------
Please review changes.
Run the tests.

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -195,7 +195,7 @@ remove_dependencies() {
             ZYPPER_OPTIONS="--non-interactive"
             # zypper version 7.4 got support for automatically removing
             # unneeded packages, but only when removing installed packages.
-            if [ $(rpm -qv libzypp) \> 'libzypp-7.4' ]; then
+            if [ $(rpm -qv libzypp) \> 'libzypp-7.39' ]; then
                 ZYPPER_OPTIONS="$ZYPPER_OPTIONS --clean-deps"
             fi
             execute sudo zypper $ZYPPER_OPTIONS remove $SLES_PACKAGES
--- a/chevah_build
+++ b/chevah_build
@@ -173,19 +173,29 @@ remove_dependencies() {
         ;;
         rhel*)
             execute sudo yum -y remove $RHEL_PACKAGES
-            # Nothing else works in all supported RHELs (currently 4 to 7)
-            # for automatically removing packages installed as dependencies.
-            RHEL_LEAVES=`package-cleanup --leaves --quiet 2>/dev/null \
-                | egrep -v ^'Excluding|Finished'`
-            if [ ! -z $RHEL_LEAVES ]; then
-                execute sudo yum -y remove $RHEL_LEAVES
+            if [ $OS \> 'rhel6' ]; then
+                execute sudo yum -y autoremove
+            else
+                # This partially works in RHEL 4 to 6 for automatically
+                # removing packages installed as dependencies.
+                rhel_yum_autoremove() {
+                    RPM_LEAVES=$(package-cleanup --leaves --quiet 2>/dev/null \
+                        | egrep -v ^'Excluding|Finished')
+                    if [ -z "$RPM_LEAVES" ]; then
+                        (exit 0)
+                    else
+                        execute sudo yum -y remove "$RPM_LEAVES"
+                        rhel_autoremove
+                    fi
+                }
+                rhel_yum_autoremove
             fi
         ;;
         sles*)
             ZYPPER_OPTIONS="--non-interactive"
             # zypper version 7.4 got support for automatically removing
             # unneeded packages, but only when removing installed packages.
-            if [ `rpm -qv libzypp` \> 'libzypp-7.4' ]; then
+            if [ $(rpm -qv libzypp) \> 'libzypp-7.4' ]; then
                 ZYPPER_OPTIONS="$ZYPPER_OPTIONS --clean-deps"
             fi
             execute sudo zypper $ZYPPER_OPTIONS remove $SLES_PACKAGES
--- a/chevah_build
+++ b/chevah_build
@@ -15,6 +15,8 @@
 UBUNTU_PACKAGES="gcc libssl-dev zlib1g-dev libbz2-dev m4"
 RHEL_PACKAGES="gcc openssl-devel zlib-devel bzip2-devel m4"
 SLES_PACKAGES="gcc openssl-devel zlib-devel libbz2-devel m4"
+# Currently, we use Debian 7 as a generic Linux distro.
+LINUX_PACKAGES="gcc libssl-dev zlib1g-dev libbz2-dev m4"
 
 LIBFFI_VERSION=3.2.1
 GMP_VERSION=6.0.0
@@ -139,6 +141,10 @@ install_dependencies() {
             packages=$SLES_PACKAGES
             install_command='sudo zypper --non-interactive install -l'
         ;;
+        linux*)
+            packages=$LINUX_PACKAGES
+            install_command='sudo apt-get install -y'
+        ;;
         aix*)
             # For the moment, we don't install anything on AIX.
             packages=""
@@ -180,6 +186,9 @@ remove_dependencies() {
         ;;
         sles*)
             execute sudo zypper --non-interactive remove $SLES_PACKAGES
+        ;;
+        linux*)
+            execute sudo apt-get remove -y --purge $LINUX_PACKAGES
         ;;
     esac
 }
--- a/chevah_build
+++ b/chevah_build
@@ -203,6 +203,9 @@ help_text_build="Create the Python binaries for current OS."
 command_build() {
     install_dependencies
 
+    # Clean the install sub-directory to avoid contamination.
+    rm -rf ${INSTALL_FOLDER}/*
+
     case $OS in
         aix*)
             build 'libffi' "libffi-$LIBFFI_VERSION" ${PYTHON_BUILD_FOLDER}
--- a/chevah_build
+++ b/chevah_build
@@ -342,8 +342,10 @@ help_text_test=\
 "Run a quick test for the Python from build."
 command_test() {
     test_file='test_python_binary_dist.py'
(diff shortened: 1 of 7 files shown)


chevah_build
```diff
@@ -15,6 +15,8 @@
 UBUNTU_PACKAGES="gcc libssl-dev zlib1g-dev libbz2-dev m4"
 RHEL_PACKAGES="gcc openssl-devel zlib-devel bzip2-devel m4"
 SLES_PACKAGES="gcc openssl-devel zlib-devel libbz2-devel m4"
+# Currently, we use Debian 7 as a generic Linux distro.
+LINUX_PACKAGES="gcc libssl-dev zlib1g-dev libbz2-dev m4"
```
adiroiban: this looks similar to Ubuntu ... the LINUX_PACKAGES are in fact debian package which are similar to Ubuntu..

We are doing all work on Debian, just that we can the final product linux...

chevah_build
```diff
@@ -139,6 +141,10 @@ install_dependencies() {
             packages=$SLES_PACKAGES
             install_command='sudo zypper --non-interactive install -l'
         ;;
+        linux*)
```
adiroiban: maybe we can have
```
ubuntu*|linux*)
     DO_SOMETHING
```

dumol:
**needs-review**

python-modules/chevah-python-test/get_binaries_deps.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
```
adiroiban: we need a few words describing the use/purpose of this file and how it should be used/called.

Why do we need this bash script and not implement it in Python?

chevah_build
```diff
@@ -194,6 +213,9 @@ help_text_build="Create the Python binaries for current OS."
 command_build() {
     install_dependencies
 
+    # Clean the build sub-directory to avoid contamination.
```
adiroiban: contamination from what? :) ... maybe from previous builds ?

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -3,106 +3,274 @@
...
+            'libgcc_s.1.dylib',
+            'libssl.0.9.8.dylib',
+            'libz.1.dylib',
+            ]
+    return expected_deps
+
+
+def get_actual_deps():
+    """
+    Here we get the list of actual deps using a shell script helper.
```
adiroiban: Avoid "Here to DO somehting" from doctrings. Keep it simple. Don't add details about implementation.
The docstring is there for the people using this method.
It should state  what it does and the format of the returned value.

"""
Return a list with path of dependent libraries.

Path can be relative or absolute.
"""

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -3,106 +3,274 @@
...
+    """
+    global exit_code
```
adiroiban: we should avoid global.

You can return `None` on errors and then the caller can set wather exit_code it likes.

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -3,106 +3,274 @@
...
+    try:
+        actual_deps = subprocess.check_output('./get_binaries_deps.sh',
```
adiroiban: bad indentation. see styleguide

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -3,106 +3,274 @@
...
+                      shell=True).split()
+    except:
+        print 'Could not determine the deps for the new binaries.'
```
adiroiban: maybe this should be printed to stderr rather than stdout ?

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -3,106 +3,274 @@
...
+
+
+def check_deps(expected_deps, actual_deps):
+    """
+    We check actual deps one by one to see if there is a substring of each
+    of them in any dep from the list of expected deps. This is so that an
+    actual dep of libssl.so.0.9.8 or libssl.so.1.0.0 matches an expected dep of
+    libssl.so when checking for OpenSSL.
+    """
+    global exit_code
```
adiroiban: global are bad...

you can return True or False if deps are valid or not.

You can rename `check_deps` into `are_deps_valid` to make it clear that it will return a boolean value and should be used in a test.

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -3,106 +3,274 @@
...
+        for single_dep_to_print in unwanted_deps:
+            print '\t' , single_dep_to_print
+        exit_code = 15
+
+
+def main():
+    """
+    Main function.
+    """
+    global exit_code
```
adiroiban: we don't need this

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -3,106 +3,274 @@
...
+    if unwanted_deps:
+        print 'Got unwanted deps:'
+        for single_dep_to_print in unwanted_deps:
+            print '\t' , single_dep_to_print
+        exit_code = 15
+
+
+def main():
+    """
+    Main function.
```
adiroiban: this comment does not add any value... as `def main` already state that.

You should document what the function does. ie, launch test to validate modules and dependencies

adiroiban:
Thanks for the changes
This is fine, but the current diff does much more than providing support
for generic linux x86

The title of this review is "Generic Linux x86 support." but the changes do:

* Refactors the way packages are installed/removed on non linux x86
* add a test for dependencies
* general refactoring in paver.sh

Is ok to have drive-by fixes, but in current code it looks like adding
linux x86 support is the drive by fix .. and the other changes are huge,
and hard to be described as drive-by :)

It looks like an approptiate title for this review would be "My work for the last week" :)
--------

I know that I have approved the previous changes for paver.sh but I would
like to keep paver.sh as close to a minimal as possible, with only
helpers for runtime.

I would like to avoid growing it into a monster which does everything.

The detect_os part should do just that, detect the OS/CPU.

I know that is also detect the python version and build folder, but those
are there for legacy reasons and we should remove them and move them in
paver.conf... python version was already moved in paver.conf

DEFAULT_VALUES files is a bad design and is kept for backward compatibility
reasons.

Compile-time helpers should go in other script which will reside only in
python-package

-----

My suggest is to split this code into the following tickets:

* in brink: Refactors the way packages are installed/removed on non linux x86
* in brink: Refactoring of paver.sh os detection
* here: Add test for dependecies
* here: Support for linux x86

This review is not a big deal but we should develop good habits and create
sound changes which focus on a specific thing.

needs-changes

dumol:
Thanks for the review!I know I crossed a line with so many drive-by fixes, but it's complicated… :) I'll try to address first the issues with `paver.sh` so that I know what path to take.

Most of the brink-related changes are already in the master branch of the brink repo. And in compat, watchdog etc. Unfortunately I can't merge with brink's master to only show the newer additions because here I'm in another repo.

I can't reset to `paver.sh` in python-package's either, because I need an updated `paver.sh` for the core reason of this issue, which is to support generic Linux installations.

Also, one of the changes you seem to suggest I should revert are already in brink's master. I am referring here to extending DEFAULT_VALUES with CC and CXX.

A solution would be to drop this altogether and start from scratch. First with brink (two or more issues, as you requested) and then with python-package (importing changes in brink, supporting generic Linux, checking deps etc.).

Another solution would be to backport the relevant changes to brink (they are minor) and review them only there. And only review non-brink changes here.

But fist please enlighen me in regards to the DEFAULT_VALUES in `paver.sh`, because extending it has passed review in brink and this is in master already.

Thus, for additional feedback:

**needs-review**

adiroiban:
> But fist please enlighen me in regards to the DEFAULT_VALUES in paver.sh, because extending it has passed review in brink and this is in master already.

True. My bad. For now we will keep CC handling in DEFAULT_VALUES as it is.
I was just a comment for the general direction of paver.sh

-----

> Most of the brink-related changes are already in the master branch of the brink repo. And in compat, watchdog etc. Unfortunately I can't merge with brink's master to only show the newer additions because here I'm in another repo.
> 
> I can't reset to paver.sh in python-package's either, because I need an updated paver.sh for the core reason of this issue, which is to support generic Linux installations.

OK. My bad again.

There were some many commits here and I found it hard to see which changes to paver.sh are just a sync with latest version from brink... I have also forgot about the sync.

In case the changes in paver.sh are just due to applying updated from brink, than they are ok.

----------

In case paver.sh changes are already in brink then brink is fine. You will need only 2 more branches.

* python-package for chevah_build - Refactors the way packages are installed/removed on non linux x86
* python-package: Add test for dependecies
* python-package: Support for linux x86 - this branch

thanks!

needs-changes

dumol:
All the improvements in this branch have been implemented in #16, #17, #18 and #19. So, it seems the pull #13 was indeed unlucky! :D

Closing…

PR #18, opened 2015-02-09
[#2680] Check unwanted deps.

dumol:
Problem?
-------------
While working on #2647 a need arose to check the dependencies for the new Python binaries in `build/` so that our packages don't get contaminated with unwanted deps.

Solution?
------------
Use `ldd` or equivalents to build a list of deps for the build and compare it with a hardcoded minimal list for every supported OS.

**Drive-by fix**:
  * disable a few more modules to minimize dependencies for new builds.

How to check?
--------------------
Please review changes.
Run the tests.

reviewer: @adiroiban

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -355,8 +355,10 @@ help_text_test=\
 "Run a quick test for the Python from build."
 command_test() {
     test_file='test_python_binary_dist.py'
+    test_file_helper='get_binaries_deps.sh'
     execute mkdir -p build/
     execute cp python-modules/chevah-python-test/${test_file} build/
+    execute cp python-modules/chevah-python-test/${test_file_helper} build/
     execute pushd build
     execute ./$LOCAL_PYTHON_BINARY_DIST/bin/python ${test_file}
     execute popd
--- a/chevah_build
+++ b/chevah_build
@@ -355,10 +355,9 @@ help_text_test=\
 "Run a quick test for the Python from build."
 command_test() {
     test_file='test_python_binary_dist.py'
-    test_file_helper='get_binaries_deps.sh'
     execute mkdir -p build/
     execute cp python-modules/chevah-python-test/${test_file} build/
-    execute cp python-modules/chevah-python-test/${test_file_helper} build/
+    execute cp python-modules/chevah-python-test/get_binaries_deps.sh build/
     execute pushd build
     execute ./$LOCAL_PYTHON_BINARY_DIST/bin/python ${test_file}
     execute popd
--- a/python-modules/chevah-python-test/get_binaries_deps.sh
+++ b/python-modules/chevah-python-test/get_binaries_deps.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+checker=ldd
+os=$(uname)
+
+# In OS X, there's no ldd.
+if [ "$os" = "Darwin" ]; then
+    checker="otool -L"
+fi
+
+# In Solaris 10, make sure we find the OpenSSL libs (both 64/32 binaries).
+if [ "$os" = "SunOS" ]; then
+    # We need a trick to avoid "unbound variables" errors with "set -o nounset".
+    # https://www.gnu.org/software/bash/manual/bashref.html#Shell-Parameter-Expansion
+    export LD_LIBRARY_PATH="${LD_LIBRARY_PATH:+$LD_LIBRARY_PATH:}/usr/sfw/lib/64:/usr/sfw/lib/"
+fi
+
+# This portable invocation of find will get a raw list of linked libs
+# for the current binaries in the current sub-directory: 'build'.
+raw_list=$(find ./ -type f \( -name "python" -o -name "*.so" \) \
+            -exec $checker {} \;)
+
+# We exclude any occuring lines listing the examined files as we
+# know they start with './' and we select the first fields with awk.
+lib_list=$(echo "$raw_list" | grep -v ^\\./ | awk '{print $1}')
+
+# For AIX and OS X, the output includes the full path. More so, some
+# AIX 5.x libs from /usr/lib/ have moved to /lib in newer AIX versions.
+if [ "$os" = "AIX" ]; then
+    >&2 echo -n "Beware that in AIX we cut /lib/ and /usr/lib/ prefixes from the "
+    >&2 echo    "full paths to the libs in the list."
+    lib_list=$(echo "$lib_list" | sed s#^/usr##g | sed s#^/lib/##g \
+                | cut -d \( -f1)
+elif [ "$os" = "Darwin" ]; then
+    >&2 echo -n "Beware that in OS X we cut /System/Library/Frameworks/ and "
+    >&2 echo    "/usr/lib/ prefixes from the full paths to the libs in the list."
+    lib_list=$(echo "$lib_list" | sed s#/System/Library/Frameworks/##g \
+                | sed s#^/usr/lib/##g)
+fi
+
+# Finally, we sort the list to remove duplicates.
+echo "$lib_list" | sort | uniq
--- a/python-modules/chevah-python-test/get_binaries_deps.sh
+++ b/python-modules/chevah-python-test/get_binaries_deps.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-# Outputs a sorted list of dependencies for the python binary and .so files
-# in the current directory and its subdirectories (to be run in 'build/').
+# Uses 'ldd' or equivalents to list all dependencies for the python binary and
+# .so files in the current hierarchy of directories (to be run in 'build/').
 
 set -o nounset
 set -o errexit
@@ -22,5 +22,5 @@ if [ "$os" = "SunOS" ]; then
 fi
 
 # This portable invocation of find will get a raw list of linked libs
-# for the current binaries in the current sub-directory: 'build'.
+# for the current binaries in the current sub-directory, usually 'build'.
 find ./ -type f \( -name "python" -o -name "*.so" \) -exec $checker {} \;
--- a/python-modules/chevah-python-test/get_binaries_deps.sh
+++ b/python-modules/chevah-python-test/get_binaries_deps.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+# Outputs a sorted list of dependencies for the python binary and .so files
+# in the current directory and its subdirectories (to be run in 'build/').
+
 set -o nounset
 set -o errexit
 set -o pipefail
--- a/python-modules/chevah-python-test/get_binaries_deps.sh
+++ b/python-modules/chevah-python-test/get_binaries_deps.sh
@@ -10,7 +10,6 @@ set -o pipefail
 checker=ldd
 os=$(uname)
 
-# In OS X, there's no ldd.
 if [ "$os" = "Darwin" ]; then
     checker="otool -L"
 fi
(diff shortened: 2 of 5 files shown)


dumol:
Implemented most suggestions from https://github.com/chevah/python-package/pull/13 with the exception of printing to stderr when the list of deps is empty. I think that, if we want that, then we should do it for all the cases when we subsequently set `exit_code` to something other than 0. Should I do that?

Also, an explanation for the bash script helper… I'm not aware of how we could determine dependencies for the binaries with our minimal Python build so I think we need to run at least an external command for that. I chose to use a helper script to build a list of deps, because the Unix-style utilities are better catered to that: faster to write a script to parse the output of find, faster to run etc. I've read once an article on this from Rob Pike, I think, comparing processing logs in C or with basic Unix utilities. Can't find it now, unfortunately…

dumol:
**needs-review**

chevah_build
```diff
@@ -355,8 +355,10 @@ help_text_test=\
 "Run a quick test for the Python from build."
 command_test() {
     test_file='test_python_binary_dist.py'
+    test_file_helper='get_binaries_deps.sh'
```
adiroiban: test_file_helper ? I think that this is a bad name for this variable.... since this is used in a single place, do we need a variable for it?

python-modules/chevah-python-test/get_binaries_deps.sh
```diff
@@ -0,0 +1,49 @@
...
+# in the current directory and its subdirectories (to be run in 'build/').
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+checker=ldd
+os=$(uname)
+
+# In OS X, there's no ldd.
```
adiroiban: not sure if this comment add any new info :)

python-modules/chevah-python-test/get_binaries_deps.sh
```diff
@@ -0,0 +1,49 @@
...
+fi
+
+# This portable invocation of find will get a raw list of linked libs
+# for the current binaries in the current sub-directory: 'build'.
+raw_list=$(find ./ -type f \( -name "python" -o -name "*.so" \) \
+            -exec $checker {} \;)
+
+# We exclude any occuring lines listing the examined files as we
+# know they start with './' and we select the first fields with awk.
+lib_list=$(echo "$raw_list" | grep -v ^\\./ | awk '{print $1}')
```
adiroiban: I would prefer that this script just return a raw list and filter it later in python.... we can also sort / unique filter in python.

i don't like this echo | sed | sed | cut chaining as I find it hard to read and maintain

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -3,99 +3,278 @@
...
+            shell=True).split()
+    except:
+        print 'Could not determine the deps for the new binaries.'
+    else:
+        return actual_deps
+
+
+def check_deps(expected_deps, actual_deps):
+    """
+    Returns unwanted deps for the python binary or the .so files in 'build/'.
```
adiroiban: I don't understand the last part of this docstring. ... " or the .so files in build/" 

if it returns unwanted deps, then method should be named `get_unknown_dependent_libraries` or something like this.

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -3,99 +3,278 @@
...
+    """
+    # Check actual deps one by one to see if there is a substring of each of
```
adiroiban: This comment should be part of the docstring as it describes what expected_deps and actual_deps are and what they should contain.

python-modules/chevah-python-test/test_python_binary_dist.py
adiroiban: also you should specify that it returns a list of not filtered deps

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -3,99 +3,278 @@
...
+        try:
+            import spwd
+            spwd
+        except:
+            print 'spwd missing.'
+            exit_code = 1
+
+    # Finally, compare the list of expected deps for the current OS with the
+    # list of actual deps returned by the shell script helper.
+    expected_deps = set_expected_deps()
```
adiroiban: set_expected_deps or get_expected_deps ? or get_allowed_deps or something like this

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -3,99 +3,278 @@
...
-    _fastmath
-except:
-    print 'Crypto.PublicKey._fastmath missing. No GMP?'
-    exit_code = 1
-
-# Windows specific modules.
-if os.name == 'nt':
+
+
+def set_expected_deps():
```
adiroiban: method is called set_ but in fact it gets.

it does not return a list for every supported os. but the list of approved deps for **current** os

adiroiban:
Changes look good. Thanks!

We need to push more logic as Python code.... I don't like "shell sausages"

They need a bit of cleanup before merge so that next time when we read the code we don't have WTF moments... http://www.osnews.com/story/19266/WTFs_m

needs-changes

python-modules/chevah-python-test/test_python_binary_dist.py
dumol: What do you mean _not filtered_? `expected_deps` is hardcoded and `actual_deps` is already filtered.

python-modules/chevah-python-test/test_python_binary_dist.py
dumol: I'm a bit confused. In https://github.com/chevah/python-package/pull/13 you said: _[…] doctrings. Keep it simple. Don't add details about implementation. The docstring is there for the people using this method. It should state what it does and the format of the returned value_.

Besides, the comment doesn't actually describe what `expected_deps` and `actual_deps` are and what they should contain. Or there is something evading my attention?

So what should I do? Add what  what `expected_deps` and `actual_deps` are and what they should contain in the docstring? From PEP 257 I gather this is required. But then, it enforces things you haven't asked for, so I'm not sure.

What about the the 4-lines comment below that explains the logic checking the deps? Add that too to the docstring?. I have no opinion on that whatsoever and looking to the style guide and PEPs I'm not sure.

dumol:
Implemented most of the requested changes. Where in doubt, I have commented above (in the outdated diff). Also, please address my first comment, still visible after the first two commits. Thanks!

**needs-review**

dumol:
I also wonder about this:
```
# Copyright (c) 2011 Adi Roiban.
# See LICENSE for details.
```
Should I update it? If yes, how? :)

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -3,99 +3,285 @@
...
+        raw_deps = subprocess.check_output(script_helper).splitlines()
     except:
-        print '"ctypes - windll" missing.'
-        exit_code = 1
+        sys.stderr.write('Could not get the deps for the new binaries.\n')
+    else:
+        libs_deps = []
+        for line in raw_deps:
+            # We exclude any occuring lines listing the examined files as we
+            # know they start with './' (this is a problem in AIX and OS X).
```
adiroiban: If this is a problem on AIX/OSX maybe we should only ignore them on those platforms

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -3,99 +3,285 @@
...
+            if line[:2] != './' :
```
adiroiban: this look like `if not line.startswith('./'):`  I find it much easier to read

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -3,99 +3,285 @@
...
+        return list(set(libs_deps))
+
+
+def get_unwanted_deps(allowed_deps, actual_deps):
+    """
+    Return unwanted deps for the newly-built binaries.
+    allowed_deps are hardcoded deps for binaries built for the current OS.
+    actual_deps are the deps found for the newly-built binaries.
+    """
+    # Check actual deps one by one to see if there is a substring of each of
```
adiroiban: this should be merged into the docstring as it is a generic comment valid for the whole function.

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -3,99 +3,285 @@
...
+            spwd
+        except:
+            sys.stderr.write('spwd missing.\n')
+            exit_code = 12
+
+    # Compare the list of allowed deps for the current OS with the list of
+    # actual deps for the newly-built binaries returned by the script helper.
+    allowed_deps = get_allowed_deps()
+    if not allowed_deps:
+        sys.stderr.write('Got no allowed deps. Please check if {0} is a ' \
```
adiroiban: don't use '\' for spliting long lines. you can splint string over long lines without any problems
```
sys.stderr.write(
    'Got no allowed deps. Please check if {0} is a '
    'supported operating system.\n'.format(platform.system()))

**Comment by adiroiban** [comment]
Changes looks good. Just minor comments ...for this stage the code is ok.

I will try to make you a lean introduction to our coding standards and best practices in writing maintanable code.

------

For the copyring... I have no idea... leave it that way.
I think that for the future we should remove the  year and put something like Chevah Team
....

for the chevah-python repo we need to LICENSE file to describe the license for our scritps.

I will go with BSD as the single license used in our open source projects.

Thanks!

approved-at latest

File: python-modules/chevah-python-test/test_python_binary_dist.py (hunk 8 above)
Comment (dumol) [inline python-modules/chevah-python-test/test_python_binary_dist.py]: I tried this suggestion but things looked unnecessarily complicated. What do we gain from this? If we want speed we should use `sort` and `uniq`, not Python… Besides, I wouldn't be surprised if other OS'es have the same problem (BSD and maybe HP-UX). So, for the moment, I have stashed those changes.

For later reference:
```
         for line in raw_deps:
-            # We exclude any occuring lines listing the examined files as we
-            # know they start with './' (this is a problem in AIX and OS X).
-            if line[:2] != './' :
+            if ( platform_system == 'aix' ) or ( platform_system == 'darwin' ):
+                # We exclude any occuring lines listing the examined files as
+                # we know they start with './' (a problem in AIX and OS X).
+                if line[:2] != './' :
+                    libs_deps.append(line.split()[0])
+            else:
                 libs_deps.append(line.split()[0])
         return list(set(libs_deps))
```

File: python-modules/chevah-python-test/test_python_binary_dist.py (hunk 9 above)
Comment (dumol) [inline python-modules/chevah-python-test/test_python_binary_dist.py]: Thanks for the suggestion!

File: python-modules/chevah-python-test/test_python_binary_dist.py (hunk 10 above)
Comment (dumol) [inline python-modules/chevah-python-test/test_python_binary_dist.py]: As said above, I'm a bit confused. In #13 you said: _[…] doctrings. Keep it simple. Don't add details about implementation. The docstring is there for the people using this method. It should state what it does and the format of the returned value._

I have no opinion on that whatsoever, but by looking to the style guide and PEPs (especially PEP 257), I think I shouldn't add that comment. Following those docs, I have added only the lines for the return value and the parameters. But then, the PEPs enforce things you haven't asked for, so I'm not sure.

Please clarify this for me. Thanks!

File: python-modules/chevah-python-test/test_python_binary_dist.py (hunk 11 above)
Comment (dumol) [inline python-modules/chevah-python-test/test_python_binary_dist.py]: Thanks!

File: python-modules/chevah-python-test/test_python_binary_dist.py (hunk 9 above)
Comment (adiroiban) [inline python-modules/chevah-python-test/test_python_binary_dist.py]: A much better approach is... we are in python ... not ASM, C or primitive shell, try not to think in terms of low level operation, but rather in high level operations.

```
for line in lines:
    if line.startswith('./'):
        # Reason whey we ignore this
        continue
    # explain why we need to split this
    deps = line.split()[0]
    libs_deps.append(deps)
```

python-modules/chevah-python-test/test_python_binary_dist.py
adiroiban: this is not about speed as this is only called once.

the idea is that we don't accidentally filter bad paths on linux 

but maybe you should improve the comment... as I have no idea what is going on here :)

dumol:
A couple of minor clarifications needed, please see the discussions above. As you seem to ignore my comments otherwise, I'll ask for a new review… :)

Also, I'll reiterate two notes from the very first comment to this PR:

  * in addition to #13, I have added a few more entries to the list of disabled modules so that tests pass in Gentoo and Slackware, where there are lots of header files installed by default and I end up with unwanted deps to the `readline`, `gdbm` and `ncurses` OS libs. This build should now always end up with minimal dependencies, no matter what headers are installed. Are you OK with this?

  * an interesting thing to note: in AIX the build required the zlib-devel RPM from AIX for the zlib module, but not the zlib RPM, because our binaries get linked to the OS zlib libs. Is this fine?

**needs-review**

python-modules/chevah-python-test/test_python_binary_dist.py
dumol: Fine by me… I'll also try to rephrase the comment to make it clear what is this all about.

adiroiban:
> in addition to #13, I have added a few more entries to the list of disabled modules so that tests pass in Gentoo and Slackware, where there are lots of header files installed by default and I end up with unwanted deps to the readline, gdbm and ncurses OS libs. This build should now always end up with minimal dependencies, no matter what headers are installed. Are you OK with this?

yes. This is OK. ... it would be nice to have support for readline, but I prefer to add it later as a separate task and as a separate module.


> an interesting thing to note: in AIX the build required the zlib-devel RPM from AIX for the zlib module, but not the zlib RPM, because our binaries get linked to the OS zlib libs. Is this fine?

yes. I assume we use the zlib-devel RPM from here http://www-03.ibm.com/systems/power/software/aix/linux/toolbox/date.html so it should be similar to the zlib provided by AIX... just packed as RPM 

it might be a problems since zlib-devel RPM might be of a different version than the AIX zlib, but zlib headers/API is stable and there is a low risk of getting into troubles.

Thanks!

approved-at latest

PR #20, opened 2015-02-16
[#2700] Use updated buildbot with bz2file.

dumol:
Problem?
-----------
With the new python-package, which has no `bz2` module, builtbot fails because it needs `bz2`.

Solution?
----------
Updated `pavement.py` to use a patched buildbot that uses `bz2file`.

How to test?
--------------
Please review changes.
Run tests.

reviewer: @adiroiban

### DIFF
--- a/pavement.py
+++ b/pavement.py
@@ -7,7 +7,6 @@ Build script for Python binary distribution.
 RUN_PACKAGES = [
     'zope.interface==3.8.0',
     'twisted==12.1.0-chevah3',
-    'bz2file==0.98',
 
     # Buildbot is used for try scheduler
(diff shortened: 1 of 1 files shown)


dumol:
**needs-review**

pavement.py
```diff
@@ -7,9 +7,10 @@
 RUN_PACKAGES = [
     'zope.interface==3.8.0',
     'twisted==12.1.0-chevah3',
+    'bz2file==0.98',
```
adiroiban: this is no longer needed

adiroiban:
changes looks good. see comment

approved-at latest

PR #22, opened 2015-02-20
[#2700] Use updated buildbot without bz2.

dumol:
Problem?
-----------
With the new python-package, which has no `bz2` module, builtbot fails because it needs `bz2`.

Solution?
----------
Updated `pavement.py` to use a patched buildbot that doesn't need `bz2`.

**Note**: This pull request is a another attempt at this, as #20 somehow got closed and its branch deleted both here and on our repo without merging to master. No idea how that happened, I was still at `test_review`. Couldn't determine from the git-hooks' output log either.

How to test?
--------------
Please review changes.
Run tests.

reviewer: @adiroiban

### DIFF
--- a/pavement.py
+++ b/pavement.py
@@ -9,7 +9,7 @@ RUN_PACKAGES = [
     'twisted==12.1.0-chevah3',
 
     # Buildbot is used for try scheduler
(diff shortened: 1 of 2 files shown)


dumol:
**needs-review**

PR #23, opened 2015-02-20
[#2710] ARM64 support in Linux.

dumol:
Problem?
-----------
No support for ARM64 Linux.

Solution?
-----------
Fix problems encountered on ARM64.

**Note**: the patch to Python sources was only needed for 2.7.8.

How to test?
--------------
Please review changes.
Run the tests.

reviewer: @adiroiban

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -64,6 +64,10 @@ export CC
 # when not using gcc and thus silence the associated configure warning. However,
 # we'll need to set CPPFLAGS later for linking to statically-compiled libs.
 export CXX
+# Use PIC (Position Independent Code) with GCC on 64-bit arches.
+if [ "$CC" = 'gcc' -a ${ARCH%%64} != "$ARCH" ]; then
+    export CFLAGS="${CFLAGS} -fPIC"
+fi
 
 LOCAL_PYTHON_BINARY_DIST="$PYTHON_VERSION-$OS-$ARCH"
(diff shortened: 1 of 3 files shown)


dumol:
**needs-review**

adiroiban:
Changes looks good. Please copy the ARM64 python distribution in our production location so that I can create a server distribution and see if it works.

I saw that builds are failing.

Thanks!

approved-at latest

dumol:
Builds are fine, but the package management failed on a RHEL machine. Which made me realize that adding `make` to the list of packages is not a good idea, because with the current code it gets removed at the end of the build phase.  Which renders our RHEL/SLES/Ubuntu build slaves unusable.

I'll remove the drive-by fix and come with a proper solution, that I have already tested. Which is to only remove packages that were installed during the build phase. This simple fix will allow us to add `make` to the list of installed packages without harming the build slaves that need it installed all the time. And will generally solve the annoying issue of uninstalling packages that were already installed before the build phase.

**needs-changes**

adiroiban:
ok. thanks!

dumol:
**needs-review**

adiroiban:
looks good.  looking forward for the package to be manually uploaded on binary.chevah.com so that we can start to bootstrap the buildslave and also the server.

Thanks!

approved-at  3e5fb80

PR #24, opened 2015-02-23
[#2713] Only remove installed packages.

dumol:
Problem?
-----------
There are currently two issues with the package management in the `python-package` repo:

 1. We don't install `make` although it is required. And that's because we don't want to remove it at the end of the build as we need it on our build slaves for the other tests.

 2. We remove all managed packages, although some could have been installed before-hand. Eg. we uninstall `gcc` even if it was installed before the build. This is very annoying for the unsuspecting user…

Solution?
-----------
Use a new variable to keep count of the packages installed during build and only remove those.

**Drive-by fixes**:
  * get rid of string comparisons for version checks
  * style-guide fixes in `remove_dependencies()`.

How to check?
-----------------
Please review changes.
Run the test.

reviewer: @adiroiban

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -178,6 +178,10 @@ install_dependencies() {
 # installed by `install_dependencies` and leave the system clean.
 #
 remove_dependencies() {
+    local rpm_leaves
+    local zypper_options
+    local libzypp_version
+
     case $OS in
         ubuntu*)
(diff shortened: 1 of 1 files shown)


dumol:
**needs-review**

adiroiban:
looks good. thanks

approved-at  e34341f

PR #25, opened 2015-02-23
[#2703] Add support for HP-UX

adiroiban:
# Problem

We want to build python on hpux
# Changes (adi)

For now we target 32bit builds.
The system on which this is run is supposed to have run chevah-os-boostrap script frist, ie it has gmake/gcc already installed.

I have updated the build script to use gmake 

I tried with  aCC (HP compiles) and gcc.

I was not able to build zlib support in python with aCC and gcc.

I was not able to build GMP using aCC.
When buildilng wiht GCC some GMP tests failed. We need to see if they affect us and also report the bugs.

I was not able to build libffi with aCC but it worked with gcc

With aCC (HP compiles) I was able to create python but ctypes and zlib were failing. 

# Current results (adi)

I have uploaded in production a hybrid version of Python. It was build using aCC, but the zlib  were copied from the 2.7.3 versions provided by HP as part of their Internet Express offering and ctypes was copied from the gcc build of python 2.7.8 
# How to test

WIP: on hold until we install an hpui-1131-dev container to run this build in our network.

# Changes (dumol)

Have updated Python and GMP in master to have the latest upstream fixes and merged the changes in this branch.
GMP 6.1.2 now builds with no problems, with any compiler (the included `/usr/ccs/bin/cc`, GMP or the latest HP compiler). 
For libffi there's currently no way to build with anything other than GCC, as noted upstream at https://github.com/libffi/libffi
Python 2.7.13 still needed some minor fixes, one from Python's bug tracking system and another one for compiling with non-GCC compilers. **Please do review those patches!**
For the `zlib` module, I have included the zlib sources to build them on HP-UX and possibly other exotic platforms.
The SSL-related modules are linked to the included 0.9.8 OpenSSL in `/usr/lib/hpux32`. Note that pycrypto will link to the version in `/usr/local/` if using inclusive path env vars. This is fortunately not the case for the buildbot env, but if that happens it will be caught by the deps test.
Note that we are not using C preprocessor flags for pointing to our include files any more, as this interferes with HP's linker sometimes. We now copy them to the `Include/` sub-dir of the Python's sources through the `chevahbs` script.
As a **drive-by fix**, the patches applied to the pristine Python sources are now documented in the `chevahbs` script.

How to check
----------------
Please review the changes.
Run the tests on HP-UX, eg. https://chevah.com/buildbot/builders/python-package-hpux/builds/44
Run the tests on the other stable build slaves, eg. https://chevah.com/buildbot/builders/python-package-gk-review/builds/43

reviewer: @adiroiban


chevah_build
```diff
@@ -217,16 +221,13 @@ command_build() {
     command_clean
 
     case $OS in
-        aix*)
-            build 'libffi' "libffi-$LIBFFI_VERSION" ${PYTHON_BUILD_FOLDER}
-            ;;
-        solaris*)
+        aix*solaris*|hpux*)
```
dumol: There's a `|` missing here between `aix` and `solaris`.

adiroiban:
we have something for hp-ux ... we either need to install gcc on bazna or create a container on bazna to run the dev slave..

will put this on hold for now and wait to see if customers really need this.

dumol:
**needs-review**

adiroiban:
Thanks. Excellent work!

--------------

This build is failing, I think that due to openssl ... either cryptography or pyopenssl

```
cd /srv/buildslave/runtime/build-hpux1131-ia64/slave/server-hpux/build
-bash-4.4$ ./paver.sh test normal.ftp.test_protocol:TestFTPSEProtocol.test_ftp_PROT_no_PBSZ


Starting new tests...
Using --randomly-seed=1490364697
Exit code is: -10
```

------------

For this branch I would be happy to see if it builds with pyopenssl 0.13 + openssl0.9.8 (as this is what we know that works) 

Then in a separate branch we can look at pyopenssl 16.0 and openssl 1.0.2

I think that we can merge this anyway, so that we can work with a smaller PR.

changes-approved

PR #27, opened 2015-02-24
[#2722] Update paver.sh from brink.

dumol:
Problem?
----------
OS detection is flawed for OS X 10.10 or newer and Solaris 9 and older.

Solution?
----------
Copied `paver.sh` from brink where `detect_os()` was re-factored to use a new `check_os_version()` function that was added as a correct and uniform solution to this problem. This also includes updates for Linux on ARM64, HP-UX and a fix for unreadable dirs in PATH.

How to check?
-----------------
Run the tests.

reviewer: @adiroiban


adiroiban:
looks good. thanks. please sync with brink.
approved-at latest

PR #28, opened 2015-03-05
[#2728] Improved console messages.

dumol:
Problem?
-----------
In AIX, Solaris, HP-UX and generic Linux we don't install any packages, but we still output this:
```
$ ./chevah_build build
Installing required packages...
```
Which may be followed by a long wait for removing the `build` sub-directory if there is one, without any console message. Which has the potential to freak any admin running this as a user with unrestricted `sudo` powers.

Solution?
------------
Only inform the user of installing packages when actually installing packages. Do it a slightly more detailed way showing what is going to be installed and how. Also inform of packages to be uninstalled.

**Drive-by fix**:
  * inform the user when removing previous build sub-directory. Sometimes this takes some time and there is no feedback for the user, eg. in our AIX WPARs.

How to check?
-----------------
Please review changes.
Run `./chevah_build build` on platforms when there are no packages to be installed, eg. AIX and generic Linux and also on platforms where there is something to install, eg. RHEL4 and Ubuntu 12.04
Run the tests.

reviewer: @adiroiban

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -16,7 +16,7 @@ UBUNTU_PACKAGES="gcc make libssl-dev zlib1g-dev m4 texinfo"
 RHEL_PACKAGES="gcc make openssl-devel zlib-devel m4 texinfo"
 SLES_PACKAGES="gcc make libopenssl-devel zlib-devel m4 texinfo"
 # List of OS packages requested to be installed by this script.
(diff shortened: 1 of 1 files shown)


dumol:
**needs-review**

adiroiban:
looks good. thanks. changes-approved

PR #29, opened 2015-03-08
[#2702] Remove CC and CXX from paver.sh

dumol:
Problem?
-----------
`./paver.sh` should be kept as thin as possible. Compilers are only used by python-package so we don't need to export them.

Solution?
-----------
Set up CC and CXX in `chevah_build` and remove them from `paver.sh`.

Drive-by fix:
  * had to create a _hpux_ section in `chevah_build` to move the HP-UX related comment from `paver.sh`, so I have followed the current (minimal) changes in #25 to pre-empt the merge in `chevah-build`.

How to test?
---------------
Please review changes.
Run the tests.

reviewer: @adiroiban

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -47,8 +47,6 @@ fi
 PYTHON_VERSION=`cut -d' ' -f 2 DEFAULT_VALUES`
 OS=`cut -d' ' -f 3 DEFAULT_VALUES`
 ARCH=`cut -d' ' -f 4 DEFAULT_VALUES`
-CC=`cut -d' ' -f 5 DEFAULT_VALUES`
-CXX=`cut -d' ' -f 6 DEFAULT_VALUES`
 TIMESTAMP=`date +'%Y%m%d'`
 rm DEFAULT_VALUES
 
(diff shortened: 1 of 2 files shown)


dumol:
**needs-review**

chevah_build
```diff
@@ -126,6 +131,10 @@ case $OS in
             export CFLAGS="-m64"
         fi
     ;;
+    hpux*)
+        # libffi and GMP do not compile with the HP compiler, so we use GCC.
```
adiroiban: this comments is out of place as the next line is about gmake

adiroiban:
Changes look good... just a minor comment

thanks!

changes-approved

chevah_build
dumol: I think it's in the right place as this is where we set up custom values for CC and CXX, eg. for AIX and Solaris above. However, somewhat counter-intuitively, we don't do that for HP-UX and the comment asserts that. Perhaps we should rephrase it to make this more clear? Suggestions welcome…

adiroiban: We can have a comment like this

```
# For HPUX we were not yet been able to compile libffi and GMP with the HP compiler.
```

PR #35, opened 2015-07-20
[#2689] Readline support on selected platforms.

dumol:
Problem?
-----------
We want  to have the `readline` module compiled on Linux and other platforms that have the `readline` libs installed by default.

Solution?
----------
Enable `readline` on selected white-listed platforms such as Linux and Solaris.

How to check?
-----------------
Please check the changes.
Run the `python-package` tests on all supported platforms.

reviewer @adiroiban

### DIFF
--- a/python-modules/chevah-python-test/test_python_binary_dist.py
+++ b/python-modules/chevah-python-test/test_python_binary_dist.py
@@ -31,6 +31,9 @@ def get_allowed_deps():
             'libz.so',
             'linux-gate.so',
             'linux-vdso.so',
+            'libncurses.so',
+            'libncursesw.so',
+            'libreadline.so',
             ]
         # Distro-specific deps to add. Now we may specify major versions too.
         linux_distro_name = platform.linux_distribution()[0]
--- a/python-modules/chevah-python-test/test_python_binary_dist.py
+++ b/python-modules/chevah-python-test/test_python_binary_dist.py
@@ -131,6 +131,8 @@ def get_allowed_deps():
                 'libelf.so.1',
                 'libsoftcrypto.so.1',
                 'libssl.so.1.0.0',
+                'libreadline.so.5',
+                'libncurses.so.5',
                 ])
     elif platform_system == 'darwin':
         # This is the minimum list of deps for OS X.
--- a/src/python/Python-2.7.8/setup.py
+++ b/src/python/Python-2.7.8/setup.py
@@ -46,7 +46,7 @@ disabled_module_list = [
     ]
 
 # Compile readline module only on platforms whitelisted below.
-if host_platform not in ('linux2'):
+if host_platform not in ('linux2', 'sunos5'):
     disabled_module_list.append('readline')
 
 def add_dir_to_list(dirlist, dir):
--- a/src/python/Python-2.7.8/setup.py
+++ b/src/python/Python-2.7.8/setup.py
@@ -42,9 +42,12 @@ disabled_module_list = [
     'bz2',
     'dbm',
     'gdbm',
-    'readline',
     'sunaudiodev',
     ]
+
+# Compile readline module only on platforms whitelisted below.
+if host_platform not in ('linux2'):
+    disabled_module_list.append('readline')
 
 def add_dir_to_list(dirlist, dir):
     """Add the directory 'dir' to the list 'dirlist' (at the front) if
--- a/src/python/Python-2.7.9/setup.py
+++ b/src/python/Python-2.7.9/setup.py
@@ -42,9 +42,12 @@ disabled_module_list = [
     'bz2',
     'dbm',
     'gdbm',
-    'readline',
     'sunaudiodev',
      ]
+
+# Compile readline module only on platforms whitelisted below.
+if host_platform not in ('linux2'):
+    disabled_module_list.append('readline')
 
 def add_dir_to_list(dirlist, dir):
     """Add the directory 'dir' to the list 'dirlist' (at the front) if
--- a/src/python/Python-2.7.9/setup.py
+++ b/src/python/Python-2.7.9/setup.py
@@ -46,7 +46,7 @@ disabled_module_list = [
      ]
 
 # Compile readline module only on platforms whitelisted below.
-if host_platform not in ('linux2'):
+if host_platform not in ('linux2', 'sunos5'):
     disabled_module_list.append('readline')
 
 def add_dir_to_list(dirlist, dir):


adiroiban:
Things are ugly with readline as it is GPL ... I found that there is editline ... and libedit which is the BSD version of readline.

I have also found that homebrew version of Python for OSX is compiled with support for libedit instead of libreadline.

Maybe we should look at libedit and enable it for all Unix platforms.

What do you think?

dumol:
OK, sorry for not clearing this up in Trac, where I've raised the issue of readline's license… I think I was mistaken. Because readline's GPL license would only be a problem if we would leverage the BSD-style Python license to build a commercial distribution, á la Active Python.

But as long as our Python distribution maintains the original GPL-compatible BSD-style license and has its sources available, I think it's safe to distribute Python binaries linked to a GPL lib such as readline. It doesn't really matter if our Python program are GPL-compatible or not, as long as the Python interpreter is. Our only obligation would be to have `python-package` sources available, which is fulfilled by this repo.

So, given that the license is not that problematic, I would go for the simplest solution: link to the system-available readline on platforms that have it. So far we got Linux and Solaris. That should considerably ease things for development and not give us additional headaches, which would be the case if we would statically link our Python to readline, as we do with libffi and GMP.

I'm asking for a review for the above issue, code changes are still a work in progress. Thanks!

**needs-review**

dumol:
Holy cow, it seems things are more complicated: _when the interpreter is extended to provide “bindings” to other facilities (often, but not necessarily, libraries), the interpreted program is effectively linked to the facilities it uses through these bindings. So if these facilities are released under the GPL, the interpreted program that uses them must be released in a GPL-compatible way._ From https://www.gnu.org/licenses/gpl-faq.en.html#IfInterpreterIsGPL

I am confused… How can someone release a commercial Python program for Linux without bundling a custom build of Python not linked to any GPL lib? Case in point: the Dropbox client… Maybe that's why on my Gentoo/OpenBSD machines the system Python is not linked to readline?

What do you think?

dumol:
The same on all our Linux and Solaris systems… The bundled Python is not linked to GPL libraries, so that it has no issues running non-GPL software.

**needs-changes**

dumol:
Ok, it seems I was wrong again. I was only checking the `python` binary, when in fact I should have checked the deps of a dynamically available `readline.so` module. All systems with a bundled Python  that I've checked (various Linux'es, Solaris'es and OpenBSD 5.7), which include a `readline.so` module are linked to a bundled GNU readline lib. OpenBSD even has it in its `base` system, which makes `readline` a featured item at http://www.openbsd.org/policy.html. The same holds true for FreeBSD: https://wiki.freebsd.org/GPLinBase

The only exceptions I've encountered our Solaris 10 amd64 VM (where there is no `readline` support in the bundled Python distro, for some reason, although there is one in our Solaris 10 SPARC installation) and our OS X 10.8 VM (where the `readline.so` module is linked against `libedit` which is a NetBSD project). And I think NetBSD has no `readline` in base, because it is available as a package: http://ftp.netbsd.org/pub/pkgsrc/current/pkgsrc/devel/readline/README.html. The problem with `libedit` seems to be that it is not fully-compatible and not that mature, from I read on the internets. 

All in all, I think we are safe linking our Python distribution to the system-available `readline` on systems that have it available. I'm not sure how safe it would be to use `readline` in non-GPL compatible projects, though… Do we plan that? Or we just want command-line facilities in our Python package?

**needs-review**

adiroiban:
Thanks for the update. As long as we don't distribute libreadline we should be fine :)
Now, this PR talks about readline but I also see changes related to ncurse.

My plan is to use readline only for debugging.
Later we can see if we can use readline in shell like commands ... for example client-shell https://github.com/chevah/server/pull/743

We don't use ncurse and we don't plan to use it.
We can compile Python without ncurse support.

needs-changes

dumol:
Actually, from this point of view I don't think it matters if we distribute `readline` or only link to it dynamically. The only thing that seems to matter would be if we plan to use the `readline` Python module in a non-GPL program. 

If we want to use `readline` for a script such as `client-shell`, I see two solutions:
  1. the easy way: license `client-shell` as GPL.
  2. the hard way: use a replacement for `readline`.

But I am not a lawyer, this is just my interpretation of https://www.gnu.org/licenses/gpl-faq.en.html#IfInterpreterIsGPL which says: _when the interpreter is extended to provide “bindings” to other facilities (often, but not necessarily, libraries), the interpreted program is effectively linked to the facilities it uses through these bindings. So if these facilities are released under the GPL, the interpreted program that uses them must be released in a GPL-compatible way._

In regards to linking to `ncurses`, I don't think we can get rid of that. We don't compile the `_curses` and `_curses_panel` modules, it's the `readline` module that gets linked to `libncurses`:
```
% ldd build/python2.7-linux-x64/lib/python2.7/lib-dynload/readline.so
        linux-vdso.so.1 (0x000003562469f000)
        libreadline.so.6 => /lib64/libreadline.so.6 (0x0000035624230000)
        libncursesw.so.5 => /lib64/libncursesw.so.5 (0x0000035623fbe000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x0000035623da0000)
        libc.so.6 => /lib64/libc.so.6 (0x00000356239f5000)
        libncurses.so.5 => /lib64/libncurses.so.5 (0x0000035623795000)
        libdl.so.2 => /lib64/libdl.so.2 (0x0000035623590000)
        /lib64/ld-linux-x86-64.so.2 (0x00000356246a1000)
```
Do you think there would be a way to compile `readline.so` without a `curses` library? The only alternative that I see in the `setup.py` code would be `tinfo`, which is a _termcap interface split out from ncurses_.

**needs-review**

adiroiban:
Thanks!

Let's have this version for now.

It is not that hard to use a readline replacement as the editline library provides the same API as readline.

changes-approves

dumol:
OK, let's use this for debugging and I'll put a separate ticket for switching to `libedit` for `client-shell` and the like. But it seems that `libedit` doesn't have fully equivalent functionality, which is the reason for projects such as: https://github.com/ludwigschwardt/python-gnureadline

adiroiban:
Thanks. We can also look at https://github.com/antirez/linenoise  ... but I think that readline compatiblity layer from editline should be enough.... and since it is already used by the OSX version of Python it should be well tested.

we don't need advance features... just history. line edit and tab completions . These features are provided by libedit and libnoise.

dumol:
I looked at `linenoise` too, but have only seen support for `editline` in Python's `readline` module. I have also seen a circa-1999 `editline` module for Python, to replace its `readline` module: https://www.python.org/ftp/python/contrib-09-Dec-1999/Misc/

PR #36, opened 2015-07-30
[#2917] Use libedit for readline in Linux.

dumol:
Problem?
-----------
We can't use the `readline` module in Python programs, because the `readline` libs has a GPL-only license. From https://www.gnu.org/licenses/gpl-faq.en.html#IfInterpreterIsGPL: _when the interpreter is extended to provide “bindings” to other facilities (often, but not necessarily, libraries), the interpreted program is effectively linked to the facilities it uses through these bindings. So if these facilities are released under the GPL, the interpreted program that uses them must be released in a GPL-compatible way._

Solution?
----------
Include the BSD-licensed `libedit` for the `readline` module and patch the latter to use `libedit`. For now, we only compile the `readline` module in Linux.

Drive-by fixes:
  * fixed an issue with _global_ vs. _command_ options for SUSE's zypper
  * fixed possible `libffi` issues with AIX 6.x and 7.x
  * preserve `$CFLAGS` and `$LDFLAGS` on all OS'es
  * updated/fixed various minor comments
  * removed the already-obsoleted but still unused 2.7.9 Python sources.

How to check?
----------------
Please review the changes.
Run the builds/tests for the current branch, which include a new test for `readline`.

reviewer: @adiroiban


dumol:
**needs-review**

dumol:
This is the main diff to review: https://github.com/chevah/python-package/compare/dfaf9d4ab0380050ed473680dda774d1835695d8...98635ff6fb7d4e960e3eea7069853d771b8a2374

And then the last two commits: https://github.com/chevah/python-package/compare/cf3d2ebe87eb1e387eb1653e543f213be2c088ee...b80cb44cdab3ad57ac0366337cdbd98d7c8c572c

adiroiban:
Thanks. Let's have it merge.

changes-approved

PR #37, opened 2015-08-05
[#2690] No libgcc dep on OS X.

dumol:
Problem?
-----------
Binaries in our OS X Python Package are dynamically linked to `libgcc`.

Solution?
-----------
As `gcc` on OS X 10.8 and newer is just a front-end to LLVM's `clang`, we should target it directly, and set corresponding platform-specific options accordingly.

Drive-by fixes:
  * only support Mac OS X 10.8 or newer, the versions which still have support from Apple.

How to test?
--------------
Please review changes.
Run the OS X tests.

reviewer: @adiroiban


dumol:
Sorry for all the commits from the merged branch. Please only look at the changes, which are not that substantial. Thanks!

**needs-review**

adiroiban:
Excellent work. Thanks!

changes-approved

PR #40, opened 2015-09-02
[#2965] OpenBSD support.

dumol:
Problem
----------
I was investigating migrating my systems to OpenBSD, and thought it would be nice to be able to run the server locally. 
Could still be interesting to test against OpenBSD in the light of the _Developing Software in a Hostile Environment_ paper: http://www.openbsd.org/papers/dev-sw-hostile-env.html.

Solution
----------
At first I have minimally patched Python 2.7.8 to work with LibreSSL.
After importing Python 2.7.10 from master, there was no need for Python-related changes.
pyOpenSSL needed a minor `ifndef` for `egd`, as LibreSSL doesn't support it at all.

**Drive-by fixes**:
 * agglutinate OS'es with undefined `packages`, `install_command`, and `check_command` to alleviate the need to add a new OS in the category
 * define the `bash` she-bang properly everywhere, not all OS'es have `/bin/bash`.

How to check
----------------
Please review the changes.
Run the `python-package` tests locally in OpenBSD.

reviewer: @adiroiban


dumol:
**needs-review**

adiroiban:
Changes looks good. Thanks!

Please create an openbsd slave so that we can run this on a "neutral" system and be  able to keep the openbsd distro in sync.

Looking forward for testing the server with openbsd.

changes-approved

dumol:
Thanks! Have filled a dedicated ticket for adding an OpenBSD slave.

PR #41, opened 2015-09-04
[#2962] Raspbian support.

dumol:
Problem
----------
Building `python-package` failed on our RPi because of `gmp` issues.

Solution
----------
Fix the `gmp` issue and other problems for building in Raspbian 7.x. 
Please notice that the focus changed from supporting the arch (armv7l) to supporting Raspbian, because the resulting package depends on a Raspbian-specific library, `/usr/lib/arm-linux-gnueabihf/libcofi_rpi.so`, which is an _optimised memcpy/memset implementation for ARM11 used in Raspberry Pi_.

**Drive-by fixes**:
  * allow `cache` to be a symbolic link in `.gitignore`
  * stop enumerating all supported OS'es with no package management
  * use `if grep` consistently in `paver.sh`.

How to test
--------------
Please review changes.
Build the updated package on our RPi running Raspbian 7.8.

reviewer: @adiroiban

### DIFF
--- a/python-modules/chevah-python-test/test_python_binary_dist.py
+++ b/python-modules/chevah-python-test/test_python_binary_dist.py
@@ -72,16 +72,13 @@ def get_allowed_deps():
                 'libtinfo.so.5',
                 ])
         else:
-            try:
-                os_release_dump = open("/etc/os-release")
-            except:
-                sys.stderr.write('Linux distro with no os-release in /etc?')
-            else:
-                if 'raspbian' in os_release_dump.read():
-                    allowed_deps.extend([
-                        'libcofi_rpi.so',
-                        'libgcc_s.so.1',
-                        ])
+            # Raspbian is a special case... If this file exists, we assume the
+            # distro passes the more stringent checks in paver.sh during build.
+            if os.path.isfile("/etc/rpi-issue"):
+                allowed_deps.extend([
+                    'libcofi_rpi.so',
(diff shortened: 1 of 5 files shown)


python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -72,6 +72,16 @@ def get_allowed_deps():
                 'libtinfo.so.5',
                 ])
         else:
+            try:
+                os_release_dump = open("/etc/os-release")
```
adiroiban: it is best to close a file descriptor as soon as you no longer need it... so in python you can use
```
with open('/etc/os-release', 'r') as os_releases_file:
   if 'raspbian' in os_release_dump.read():
      MORE_WORK_HERE
```

file will be auto-close at the end of the with block 

Since this is a test code I think that we can let it fail with an OS error if file does not exists

adiroiban:
all good. the build failed due to a rogue file... i have removed it and rerun the build

only minor comments regarding the python style :)

changes-approved

dumol:
I haven't asked for review yet… :-] I have some issues with this branch on my generic Linux desktop, need to look into it. Thanks for the tips though.

**needs-changes**

adiroiban:
true. my bad. sorry!

dumol:
**needs-review**

adiroiban:
Great. Thanks!

changes-approved

PR #42, opened 2015-09-18
[#2743] Stop using the DEFAULT_VALUES file.

dumol:
Problem
----------
We write OS-dependent values such as PYTHON_VERSION, OS, and ARCH into a temporary file named `DEFAULT_VALUES`.

Solution
----------
Make `detect_os` from `paver.sh` simply output them for `chevah_build` to parse into needed variables.

How to test
--------------
Please review changes.
Check that there is no `DEFAULT_VALUES` file created any more in the current dir when building.
Run the automated tests.

reviewer: @adiroiban

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 build/
 dist/
-DEFAULT_VALUES
 build-*/
 cache
--- a/chevah_build
+++ b/chevah_build
@@ -61,7 +61,10 @@ TIMESTAMP=`date +'%Y%m%d'`
 # package. This way we are able to force a 32bit build on a 64bit machine,
 # for example by exporting ARCH in paver.sh as "x86" instead of "x64" or
 # "ppc" instead of "ppc64".
-# We also use $ARCH when building the statically compiled libffi and GMP.
+# We also use $ARCH when building the statically compiled libffi and GMP,
+# thus the need to export it.
+# Calling the new Python binary when installing the dependencies in paver.sh
+# also requires this to be exported.
 export ARCH
 # Explicitly choose the C compiler in order to make it possible to switch
 # between native compilers and GCC on platforms such as AIX and Solaris.
--- a/chevah_build
+++ b/chevah_build
@@ -8,6 +8,9 @@
 # publish_staging
 #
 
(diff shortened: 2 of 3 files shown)


dumol:
**needs-review**

adiroiban:
This is almost ok.
Please note that brink will still need those values to work.

We need to explicitly export those values to the called python binary. ex

```
diff --git a/paver.sh b/paver.sh
index afcdb17..74a2c9a 100755
--- a/paver.sh
+++ b/paver.sh
@@ -535,7 +535,7 @@ done
 
 # Now that we have Python and Paver, let's call Paver from Python :)
 set +e
-${PYTHON_BIN} -c 'from paver.tasks import main; main()' "$@"
+ARCH=$ARCH ${PYTHON_BIN} -c 'from paver.tasks import main; main()' "$@"
 exit_code=$?
 set -e
 exit $exit_code
```

or use 

```
export ARCH
```

Please export the required values and merge.

We should also document that this paver.sh will only work with brink of version newer than X

Thanks!

adiroiban:
changes-approved

adiroiban:
another option is to try to change paver.sh in brink... and also update utils.py and then apply the changes to this repo.

Thanks!

dumol:
I don't get it… `ARCH` is already exported by `./chevah_build`, I needed it for `./src/gmp/chevahbs`, and I've used it in `src/libffi/chevahbs` too, for platforms (other than AIX) needing `libffi`.

Is that enough? If yes, I could add another comment to show that it is needed when calling the new Python binary. Would that suffice? Thanks.

dumol:
**needs-review**

adiroiban:
ARCH and other info needs to be exported by the paver.sh script so that the same info which was exposed by the DEFAUL_VALUE file can be found in the env vars.

You can break that info into multiple variable names and choose the names for those new vars.

Thanks!

needs-changes

dumol:
I see… I'll implement the changes in brink too, update master there first, and then come back to fix this accordingly and merge with master. Thanks!

dumol:
Closed because its not worth the trouble, details in Trac.

PR #44, opened 2015-10-05
[#3025] Python 2.7.10 for OS'es with OpenSSL 0.9.8 or newer.

dumol:
Problem
----------
Python 2.7.8 is old and full of bugs fixed upstream in newer versions.

Solution
----------
Start to use the latest Python for OS'es that support it, starting with generic Linux. The stumbling block is that Python 2.7.9 and newer ditched support for OpenSSL 0.9.7x.

How to test
--------------
Please review my commits, excluding the first one, which is an import of the pristine upstream sources.
Run the tests.

reviewer: @adiroiban


dumol:
`test_review` failed, but the error is not reproducible: http://build.chevah.com/builders/python-package-sles-11/builds/30

**needs-review**

adiroiban:
Great work. Thanks!
Changes looks good.

A few tests are failing ... but nothing major. They are caused by the new _ssl.c code in Python 2.7.9.

While we don't use the _ssl.c code in `production` code, we do use the _ssl.c as a 3rd party SSL implementation to test our own SSL implementation.

Please don't push this into production until I will fix those tests in the server branch

I have already fixed the test but would like to have the server fixed, before publishing this in production.

Thanks!

you can merge it in master as later we can publish into production from master

changes-approved

dumol:
I'll wait then, no need to hurry. Thanks!

dumol:
Change of mind. Will put these changes in master, and let you publish production from master when you're ready. Thanks!

PR #46, opened 2015-11-09
[#3099] Build cffi module together with cryptography and openssl.

dumol:
Scope
=====

We want to have latest PyOpenSSL ... which requires python cryptography, which requires python cffi.. which requires libffi

Changes
=======

Build libffi on all system.

Build python cffi with static linked libffi.

Install pyopenssl and cryptography ... and in order to install them install pip.

Pip is installed with a cached get-pip.py stored on our servers.

I have extended the tests with a minimal check... 

The cffi tests are failing on my system. Here is how I am running them https://github.com/chevah/python-cffi/blob/build-package/python_package#L78

Here are the tests for python cryptography https://github.com/chevah/cryptography/blob/wheel-builder/python_package#L80

I did some small cleanup while reading the scripts.

How to test
=========

Once the new package is created and uploaded and staging, try to use it with server.


chevah_build
```diff
@@ -71,8 +77,10 @@ LOCAL_PYTHON_BINARY_DIST="$PYTHON_VERSION-$OS-$ARCH"
 INSTALL_FOLDER=$PWD/${BUILD_FOLDER}/$LOCAL_PYTHON_BINARY_DIST
 PYTHON_BIN=$INSTALL_FOLDER/bin/python
 PYTHON_BUILD_FOLDER="$PYTHON_VERSION-$OS-$ARCH"
+PIP_BIN=$INSTALL_FOLDER/bin/pip
```
adiroiban: maybe we can define it as 

```
PIP_BIN=$PYTHON_BIN $INSTALL_FOLDER/bin/pip
```

:)

adiroiban:
we made some progress but  I think that instead of this big branch, we should break it into separate tickets/branches targeting each OS which was left without an update.

I am tracking the progress here https://github.com/chevah/python-package/issues/57

PR #47, opened 2015-11-19
[#3125] Patch pyWin32 executables to use our included MSVC redistributable.

alibotean:
Problem
----

I've forgot to patch the executables from `pyWin32` that we use to run the product as a Windows service.

Because the bundled redist was already installed on our 2003/2008 and my local machine the tests passed.

Luckly the DC BS did not have them and the test failed.

Changes
----

I've updated the script to patch them as well.

How to test
----

reviewers @adiroiban 

Distribution is already released for testing.

Our `server` tests pass with the new distribution.

### DIFF
--- a/src/python/chevahbs
+++ b/src/python/chevahbs
@@ -149,16 +149,6 @@ chevahbs_install() {
                 --extract $destination/python27.dll.embedded \
                 $destination/python27.dll
 
-            $manifest_wiper --verbose \
-                --extract $destination/Lib/site-packages/win32/pythonservice.exe.embedded \
-                $destination/Lib/site-packages/win32/pythonservice.exe
-            $manifest_wiper --verbose \
-                --extract $destination/Lib/site-packages/win32/perfmondata.dll.embedded \
-                $destination/Lib/site-packages/win32/perfmondata.dll
-            $manifest_wiper --verbose \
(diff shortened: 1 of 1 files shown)


alibotean:
needs-review

adiroiban:
can we uninstall them from the BS... so that next time we detect these errors?

also, can we put all those version numbers in a varialbe/variables ?

changes looks good.

please check my comments and merge

thanks!

changes-approved

alibotean:
> can we uninstall them from the BS... so that next time we detect these errors?

I preferred not to as we might have other components of the system using it. If you think there is no danger I will do it.

alibotean:
> also, can we put all those version numbers in a varialbe/variables ?

I've tried at first but was unable to come up with a decent sed replacement with value so I gave up.

adiroiban:
just uninstall it... and if something fails, install it back.

thanks

alibotean:
> just uninstall it... and if something fails, install it back.

* On Windows 2003 it's not listed as a separate install (hence no uninstall option)
* Same goes for Windows 2008

I've attached two screenshots from my PC, even though I have several installed the assembly folder (which offers Uninstall on right click) does not list them :(

win7-assemblies
win7-programs

PR #49, opened 2015-12-07
[#3144] Custom malloc implementation.

dumol:
Problem
-----------
This mirrors ​https://bugs.python.org/issue3526. In short: _any memory that has been allocated by our
application at some stage will never be freed at the system level._

Solution
----------
Tried to adapt the patch at ​https://bugs.python.org/file22698/patch_dlmalloc_Python_2_7_1.diff

How to test
--------------
**WIP**
This marginally works: http://build.chevah.com/builders/python-package-linux-x86/builds/42

reviewer: @adiroiban


adiroiban:
I tried applying the patch on AIX and it is worse...

so closing this for now... but we can keep the branch available.

thanks

PR #52, opened 2016-03-23
[#3338] Use cryptography and pyOpenSSL 16.0.0 on ArchLinux builds

brunogola:
Dev environment does not work if openssl is built without SSLv3 support
=======================================================

The new builds for OpenSSL on ArchLinux are built without support for SSLv3 methods, this breaks the development environment set up because Python can't use urllib2.HTTPS* methods , as SSl.so is linked to a different OpenSSL (with support to SSLv3)

Further investigation showed that when building PyOpenSSL (0.13) or Python (2.7.10) against an OpenSSL with SSLv3 enabled it is impossible to disable those methods, so it would fail to run on systems with OpenSSL built with SSLv3 methods disabled.

Why we got into this (5 whys)
=============================

* Most distributions still build and distribute OpenSSL with SSLv3 support enabled.

Changes description
===================

* On ArchLinux build Python 2.7.10 with pyOpenSSL 16.0.0 with cryptography and cffi modules.

reviewers @adiroiban 

* check that changes make sense
* build it on other OSes and check if ```./chevah_build test```runs ok. 
* build it on ArchLinux and check if ssl, pyOpenSSL and cryptography are imported

```
$ ./chevah_build build
$ ./chevah_build test
python 2.7.10 (default, Mar 29 2016, 16:29:32) 
[GCC 5.3.0]
zlib 1.0
stdlib ssl OpenSSL 1.0.2g  1 Mar 2016
cryptography 1.3.1 - OpenSSL OpenSSL 1.0.2g  1 Mar 2016
pyopenssl 16.0.0 - OpenSSL OpenSSL 1.0.2g  1 Mar 2016
PyCrypto 2.6.1
ctypes 1.1.0
setproctitle 1.1.8-dev0
```


adiroiban:
Changes don't make sense.

For system on which we have SSLv3 we still want to support it for backward compatibility reason.

Also, please just delete the lines. We don't want to have and accumulate un-used commented code.

Thanks

adiroiban:
To me, it like your openssl version disabled SSLv3 without defining OPENSSL_NO_SSL3 ... something need to be investigate and find a better fix.

thanks

brunogola:
At compile time SSLv3 and SSLv2 are available, that's why OPENSSL_NO_SSL3 is not defined.

My build scenario on ArchLinux:

PyOpenSSL 0.13 won't compile against openssl 1.0.2 (https://github.com/pyca/pyopenssl/issues/276) so I downgraded openssl to 1.0.1l. On ArchLinux openssl 1.0.1l is compile with support to SSLv2 and SSLv3, so when I build PyOpenSSL and Python against this version of openssl, at compile time the flag OPENSSL_NO_SSL3 is not defined. 

If I run the builds after that, everything runs ok, but what I really need to test is running this builds against a newer openssl version. So I upgrade openssl to 1.0.2g, where SSLv3 is not supported. Without this patch I can't load PyOpenSSL/SSL.so or _ssl.so because they reference ssl_sslv3_method, as during compile time this was available. After the patch this works, because it will never reference ssl_sslv3_method.

The problem is: the Python and PyOpenSSL builds built against an OpenSSL with SSLv3 enabled will always need an OpenSSL distributed with those methods available as well. There is no way to build them with no support by using options during configure or build time.

If we want to build Python/PyOpenSSL without SSLv3 on a system where OpenSSL headers have no OPENSSL_NO_SSL3, then we can remove manualy this support (as I did), add our own option (NO_SSL3) or force OPENSSL_NO_SSL3 during build, as PyOpenSSL and Python have no option for that.

brunogola:
So, if we need this as an optional thing, we probably need two different builds for PyOpenSSL and Python. One with SSLv3 suppot and another without.

adiroiban:
Do we still need  `defined(NO_SSL2)` ?

I assume that we will just build python and openssl on archlinux where OPENSSL_NO_SSL2 is already defined and it will skip all this.

---------

Now we do need to check https://github.com/pyca/pyopenssl/issues/276 and see if we can update pyopenssl 0.13 to compile on latest opnenssl.


thanks!

brunogola:
I reverted my changes and now it only adds the check for OPENSSL_NO_SSL3 on PyOpenSSL.

adiroiban:
OK. So pyopenssl 0.13 is dead... we can patch it for openssl 1.0.2 but then we will need top patch it for the next release... and so on.

Also, current pyopenssl structure is very rigid and we can not have run-time detection of SSL v3 support... so if we want to work with openssl compiled without sslv3 we would need a different pyopenssl version... with a different python distro...

-------

Let's hope that python cryptography and pyopenssl 0.16 provide better support for run-time SSLv3 support detection.

Thanks!

brunogola:
As an update: Even with cryptography and pyopenssl 0.16 there is still no way to run-time SSLv3 support detection. So we need a specific build for ArchLinux. I'm working on that.

adiroiban:
That is fine. Is important to know what is possible and what not.

I am very happy to have an arch-linux build for openssl without sslv3 and sslv2 and later we  can use it in case some customer ask for it.

---------

I hope that as openssl without sslv3 became more popular a patch will be available for _ssl.c and pyopenssl to have runtime detection.

----------

The initial ticket id on trac is : Dev environment does not work if openssl is built without SSLv3 support 

Is ok to focus on having your dev environment working :)

Thanks!

brunogola:
Latest updates will build Python2.7.10 on ArchLinux with pyOpenSSL 16.0.0 and it's dependencies. 
For other OSes nothing is changed.

brunogola:
needs-review

adiroiban:
Do we still need the changed from python-modules/pyOpenSSL-0.13/OpenSSL/ssl/context.c ?

For libffi changes you should also run the build on AIX and Solaris to validate your changes.

In `test_python_binary_dist.py` DEFAULT_VALUES should be read and parsed in a separate global place to break the os detection from the actual test.

--------

why do we need this ?

```
export CHEVAH_BUILD_PATH=$INSTALL_FOLDER
```

---------

I would say that pip should be installed in python/chevahbs and not in command_build_python_extra_librarie

Also, in command_build_python_extra_libraries I would like to just have a new variable named something like EXTRA_PACKAGES which are empty for all other systems and defined for arch.
The goal is to reduce the number of `case $OS in` call in the code and have a single case $OS in at the start which sets up all the variables.

--------

PIP_ARGS is defined both in chevah_build and python/chevahbs. Can we have them defined in a single place?

I would say that windows `Installing pip + binary modules` should be done outside of python/chevahbs

and for now we can install pip for windows and archlinux.

--------------

changes looks good.
I will review them in the morning.

I would like to see a bit of less code duplication and a bit more structure in the build scripts.

The scripts are already pretty hard to read so we should see if we can improve them...just a bit.

Thanks!

brunogola:
Ah, I forgot to revert my changes to pyOpenSSL 0.13, thanks!

--------


```
export CHEVAH_BUILD_PATH=$INSTALL_FOLDER
```

is there so we can access it when building cffi (I got this from PR #46 )

--------

changes addressed in ```test_python_binary_dist.py```. I also moved the pip installation to python/chevahbs and removed some duplication and checks for $OS, but still need to check if it's on archlinux before installing pip, so I export the $OS variable. not sure if this is the best, but also this is supposed to be temporary as we might move to cryptography and pyOpenSSL 16.0.0 on other linuxes as well.

chevah_build
```diff
@@ -39,6 +40,13 @@ EXTRA_LIBRARIES="\
    python-modules/python-setproctitle-1.1.8.dev0 \
    "
 
+# We only use pip on ArchLinux and Windows builds
```
adiroiban: it should not matter at this level how we use them... soon we will use them on RHEL.. and then we will forget to update this comment and an invalid commend is worst than no comment at all.

also, a comment is a sign of 'I have no idea how to better write this code so that it will speak for itself... so here is this comment which is valid for the time when I wrote this code, but I don't guarantee that it will be valid later'

adiroiban: A better comment would be to describe how do you plan to use these variables... for ex;

```
# Space separate list of arguments sent to pip.
```

in this way someone managing them will know what type of value to put in these variables

chevah_build
```diff
@@ -59,6 +67,8 @@ TIMESTAMP=`date +'%Y%m%d'`
 # "ppc" instead of "ppc64".
 # We also use $ARCH when building the statically compiled libffi and GMP.
 export ARCH
+# We want to check if OS is archlinux when installing libffi
```
adiroiban: I think that not having this comment, and reusing the previous one should be fine

chevah_build
```diff
@@ -74,7 +84,11 @@ PYTHON_BIN=$INSTALL_FOLDER/bin/python
 PYTHON_BUILD_FOLDER="$PYTHON_VERSION-$OS-$ARCH"
 
 export MAKE=make
+
+# Used when building cffi on ArchLinux
 export CHEVAH_BUILD_PATH=$INSTALL_FOLDER
+# Used when building python on Windows or ArchLinux
+export PIP_ARGS
```
adiroiban: Do we still need to export pip_args ? I was hopping that we reduce the scope of this var to this file

src/python/chevahbs
```diff
@@ -201,6 +197,13 @@ chevahbs_install() {
             execute $MAKE install DESTDIR=$INSTALL_FOLDER
             ;;
     esac
+
+    case $OS in
+        # For now we only install PIP on ArchLinux
+        "archlinux")
+            execute $INSTALL_FOLDER/bin/python ../../python-modules/get-pip.py $PIP_ARGS
+            ;;
+    esac
```
adiroiban: I was hoping to see this moved outside of the python build.

This script should be used to build a python as closed to upstream as possible...

then in chevah_build we can do all kind of customization, like installing pip and extra libs

chevah_build
brunogola: I did not want to mess with the windows installation and PIP_ARGS is used in python/chevahbs.

adiroiban:
Is CHEVAH_BUILD_PATH really needed ?

Please see my other comments.

They are just minor.

Please address or ignore them and merge.

We can continue refactoring the script when we enable it for the other linux builds.

Thanks!

adiroiban:
changes-approved

PR #53, opened 2016-03-28
[#3381] Add DBM support in Python.

adiroiban:
# Scope

we need the 'dbm' module in our python distribution on freebsd and openbsd

At the low level that is _bsddb 
# Changes

updated tests

for now this is abadoned since we are using https://github.com/jamesls/semidbm
# How to test

WIP


adiroiban:
no longer needed... I have used a workaround for htis

PR #62, opened 2016-12-16
[#3732] Add Solaris 10 U3 detection to paver

dispera:
Scope
=====

Implement Solaris 10 U3 detection on paver.sh

Changes
=======

Added a variable to store the exact Solaris release year, and conditions that checks if the release is 'solaris10u3'. In that case, it saves that on the OS variable.

reviewers: @adiroiban @dumol 
You can check by running paver.sh detect_os on a Solaris10u3 system and checking the DEFAULT_VALUES file

### DIFF
--- a/paver.sh
+++ b/paver.sh
@@ -422,7 +422,21 @@ detect_os() {
         os_version_raw=$(uname -r | cut -d'.' -f2)
         check_os_version Solaris 10 "$os_version_raw" os_version_chevah
 
-        OS="solaris${os_version_chevah}"
+        # Solaris 10 updated the libc version between other minor updates,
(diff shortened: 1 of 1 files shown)


dispera:
Same as on the brick repo.

needs-review

adiroiban:
Changes looks good.

I think that the changes should be done together with 3725.

We can use 3732 as a ticket to remind us to update paver.sh in brink... but we might just want to update the chevah/brink README to talk that the source of paver.sh is chevah/python package.

It makes more sense to me.

So I think that this PR can be closed and the changes made as part of the work for 3725

Thanks!

needs-changes

PR #63, opened 2016-12-20
[#3725] Solaris 10 u3 support.

dispera:
Scope
=====

Implement Solaris 10u3 detection on paver.sh

Changes
=======

Added a variable to store the exact Solaris release, and a condition that checks if the release is 'solaris10u3'. In that case, it sends that to the OS variable, instead of the usual 'solaris10'.

reviewers: @adiroiban  
You can check by running 'paver.sh detect_os' on a Solaris10u3 system and checking the DEFAULT_VALUES file


dispera:
This is to continue #86 which was in Brink repo as it should be here (I will mention on chevah/brink README that the source of paver.sh is chevah/python package). 

Adding Solaris 10 U3 detection to paver.sh.
Tested on Solaris 10 U3, 10 U11, and Solaris 11.3 (and macOS 10.12, just to try that on a different platform works OK).

needs-review

adiroiban:
Thanks. It looks good.

There is little point in having solaris 10u3 detection if python-package is not able to build the python package on SOlaris 10u3

I think that the scope of this PR should be extended to include all the changes required to have a fully working pyhon-package on SOlaris 10u3 x86

In include paver.sh changes + automating all the other changes that you have manually done in order to build the package

let me know if you need help 

thanks!

needs-changes

dispera:
I managed to build the whole chevah python package now:
https://chevah.com/buildbot/builders/python-package-solaris-10u3-x86/builds/42

Tested the branch on Solaris 10u3, 10u11, 11u2 and Ubuntu 12.04, and it worked fine.

I ended up not needing the 'solaris10u3' detection I implemented as the changes work with other Solaris 10 versions. I wasn't sure if removing it, or leave it as we may need it again for 10 U3 on Sparc or x64; and may come handy if in the future some code change breaks the build for older Solaris versions and we have the detection already implemented.

There are some bits to take into account which I added as comments, like Solaris 10 U3 not coming with sqlite3, gmp, and having old zlib version which does not support the compression parameters used by the build. So they need to be installed to the default locations under /usr/local.

I did not have to use the x86 memory patch here as I used OpenSSL 0.9.7. Although it was part of my tests adding it to the script when trying OpenSSL 0.9.8, but it generates a LOT of memory allocation issues with that version. 

Readline does not compile as it is now - has libedit memory allocation issues-  but does not break the build neither. This may only happen on x86 if it is similar to OpenSSL 0.9.8 memory bug - so we will see how it goes with x64 and Sparc.

needs-review

src/python/Python-2.7.8/setup.py
```diff
@@ -803,8 +803,8 @@ def detect_modules(self):
                                libraries=math_libs) )
         # Detect SSL support for the socket module (via _ssl)
         search_for_ssl_incs_in = [
-                              '/usr/local/ssl/include',
                               '/usr/sfw/include/',
+                              '/usr/local/ssl/include',
```
adiroiban: does it matter if we change the order of these paths ?

adiroiban:
> I ended up not needing the 'solaris10u3' detection

The review request comment is longer that the `Changes` section from the PR description. Is that ok?



> There are some bits to take into account which I added as comments,

Added as comment where?

I don't see those comments in the PR

dispera:
I will update the PR Scope/Changes next time instead of putting a big comment like I did, that will be better.

Regarding the "There are some bits to take into account which I added as comments", I meant as comments in the scripts, sorry :P

The linked build failed with an exception, I imagined it would be a permission issue for my git user to have the build uploaded into the testing binaries site. 
But I actually wanted to ask you, is that why it failed on the upload step?

src/python/Python-2.7.8/setup.py
dispera: Yes. 
If we put '/usr/local/ssl/include' first, the build fails with unsupported hash types errors. It loads the headers of the new OpenSSL version from that path (0.9.8zh), instead of the headers of the one that come with Solaris 10 and are on /usr/sfw/include/ (0.9.7) which we used on chevah_build script. But.. the libraries loaded will still be from /usr/sfw/lib (0.9.7) so it fails badly.

I just re-tested to confirm it fails if we change the path order there.

adiroiban: in this case, I think that it needs an inline comment so that other will know that the order is important

dispera: Right, added the comment 👍

chevah_build
```diff
@@ -159,6 +159,13 @@ case $OS in
         export CC="cc"
         export CXX="CC"
         export MAKE="gmake"
+
+        # The package on Solaris 10 does not compile with CC, we use GCC.
```
adiroiban: why they don't compile?
 they used to work with CC and Sun Studio... this is what we had for the previous solaris 10 and solaris 11 on both x86 and sparc

chevah_build
```diff
@@ -180,13 +187,21 @@ case $OS in
             # We favour the BSD-flavoured "install" over the default one.
             # "ar", "nm" and "ld" are included by default in the same path.
             export PATH=/usr/ccs/bin/:$PATH
-            # sqlite3 lib location in all Solaris'es (incl. 10s10 for Sparc).
+            # sqlite3 lib location in all Solaris'es (incl. 10s10 for Sparc)
+            # is /usr/lib/mps.
+            # Solaris 10u3 does not come with sqlite3 or gmp, and zlib is
+            # outdated and breaks the build. If we manually install them,
```
adiroiban: why should we **manually** install them.

the whole purpose of the python-package and this chevah_build script is to automate the whole build process

adiroiban:
I think that the sqlite / libz / gmp manual compile part is wrong.

We have the sqlite / libz / gmp source code as part of this repo so that we can compile them together with python and statically / dynamically link them with our python version

we don't manually install sqlite / gmp in any other platform

also, for previous builds we managed to install everything using the Sun Studio compiler, why it is no longer working?

needs-changes

chevah_build
dispera: It failed to me like this at some points during my tests, as if trying to compile a 64bit package on a 32bit environment (note this is testing master branch on the Solaris 10 u11 x64 just now):
https://chevah.com/buildbot/builders/python-package-solaris-10/builds/94/steps/build/logs/stdio

At this moment If you test this branch with cc you would get this:
https://chevah.com/buildbot/builders/python-package-solaris-10u3-x86/builds/45/steps/build/logs/stdio

So it is only working with gcc from what I could test.

dumol:
I think it's best to not touch Diego's PR/comments and document my changes on top of his work… So, besides cleaning up the build slave (which had a lot of stuff compiled by Diego in `/usr/local`), this is what I committed:

  * merged with master, bringing ZLIB and newer GMP and SQLite in this branch
  * improved the detection for Solaris 10 releases older than u8 in `paver.sh`
  * switched to using Sun's compiler (with the exception of `libffi`, tas in HP-UX)
  * added full deps checks for all Solaris versions, so that no third-party stuff from `/usr/local/` or `/opt/` gets in our packages.

Still, there are things I don't like about this build slave, so I would keep it as _unstable_.

Please review the changes and run the tests:
https://chevah.com/buildbot/builders/python-package-solaris-10u3-x86/builds/95
https://chevah.com/buildbot/builders/python-package-group-all/builds/65

**needs-review**

adiroiban:
Thanks. I will take a look.

We have a new Sparc server and I hope that Adi L will install it soon.

For now, is ok to just merge what we have here and forget about it... for the moment the customer has not asked for 10u3 so we can ignore it

thanks

dumol:
OK, I thought this is kind of important, with one big customer and the previous release already supporting this… But no big deal, will wait for your review, wouldn't want to commit this on my own in master. Thanks!

Don't know about the Sparc server… Do you plan a 10u3 on Sparc too?

adiroiban:
Hi. Sorry. I forget the last message.

For now, I don't plan Solaris 10u3... but if the customer want to have it, we might want to provide this option.

changes-approved

PR #64, opened 2017-01-05
[#3723] Have MacOS 10.12 with homebrew OpenSSL and latest cryptography

dispera:
Scope
=====

Compile the python package on macOS 10.12 with the homebrew OpenSSL version

Why we got this
============

* OSX comes with an old OpenSSL 0.9.8 ... which does not have tls 1.2 and SHA2
* OSX uses Secure Transport API 
* Python on OSX still uses OpenSSL ... and the Apple blessed Python comes linked agains 0.9.8
* There is no Secure Transport API for Python
* Most people use homebrew OpenSSL now

Changes
=======

The OpenSSL version which comes with macOS 10.12 does not include the
headers.

So I installed OpenSSL 1.0.2j with Homebrew, which is what others recommend,
like Mozilla, and the easiest way to tell a customer to upgrade it is with brew.

After that, I just added the LDFLAGS and CPPFLAGS, and ssl_incs/libs to
our scripts, and the package was built fine on macOS 10.12:

The OS signature was updated to macos... this also signals that we will no longer use the OpenSSL provided by the OS.

With an updated OpenSSL we can also cffi and an updated python cryptography

========================

check that changes make sense

reviewers: @adiroiban


dispera:
needs-review

chevah_build
```diff
@@ -202,6 +202,9 @@ case $OS in
         export CXX="clang++"
         export CFLAGS="$CFLAGS -mmacosx-version-min=10.8"
         export MACOSX_DEPLOYMENT_TARGET=10.8
+        # The location of the Homebrew OpenSSL libs and include directories.
+        export LDFLAGS="$LDFLAGS -L/usr/local/opt/openssl/lib"
```
adiroiban: is this still for 10.8 ?

Why homebrew and not the OpenSSL version provided by OSX?

src/python/Python-2.7.10/setup.py
```diff
@@ -807,7 +807,8 @@ def detect_modules(self):
         search_for_ssl_incs_in = [
                               '/usr/local/ssl/include',
                               '/usr/sfw/include/',
-                              '/usr/contrib/ssl/include/'
+                              '/usr/contrib/ssl/include/',
+                              '/usr/local/opt/openssl/include'
```
adiroiban: I know that this is how the current code is, but I think that for new changes we need a comment to describe why we need each separate path... so if later we refactor something or we remove a supported platform we will know that is safe to remove a path

adiroiban:
> Neither the source if you download it from the OpenSSL website.

Are you sure?

How come that the source download does not include the headers?

How can the source package be used without the header files?

----------

> So I installed OpenSSL 1.0.2j with Homebrew, which is what others recommend,
> like Mozilla, and the easiest way to tell a customer to upgrade it is with brew.

If this is the way we go for OSX 10.2 we need this documented in the repo... and not only as a comment in the PR.

if we go for external homebrew, then we need dedicate documentation in SFTPPlus to instruct users how to install OpenSSL from homebrew... and also how to upgrade it in the case of security issues.

-----------

Finally we need to run this branch against the osx 10.2 slave

Looking to see how we can have the osx 10.2 slave up and running

Thanks!

chevah_build
dispera: It compiles fine on 10.8, I tested on the builder as I mentioned.

Apple removed the headers from the OS at some release, so I used Homebrew which will be the easiest way for customers.

src/python/Python-2.7.10/setup.py
dispera: Sure, I can add that.

dispera:
Scratch that about the headers not coming on the source, it seems at one point I looked at the source in the include folder which is empty, they come into ssl/ folder.

Anyway, my point was that Apple at some point (I thon 10.10) deprecated use of OpenSSL and uses its own TLS libraries for crypto, so it does not include the ssl headers with the OS.

I added the HomeBrew info to Trac, I mentioned it on the trac ticket related to this PR, but not here, look on 10.12 Sierra Section:
https://trac.chevah.com/wiki/Infrastructure/OS/OSX

dispera:
By the way, the slave is already up and running, see:
https://trac.chevah.com/ticket/3722

Just needs the builders on buildbot.

adiroiban:
In 3722 I can see that the OS name is macos-1012 ... but in this PR I don't see any changes in the paver.sh and the os is still handles as osx-108 . I that ok?

---------------

> its own TLS libraries for crypto

Is this their own OpenSSL build / libre SSL or thei reivented the wheel and have created their own SSL/TLS library?

Do you have a link in which this is discussed. 
Do you know if the change was done in 10.10 or 10.09 or 10.11?

If we plan to support OSX 10.12 we need to know what is going on with this version and to understand how OpenSSL is handled 

----------

I can see 

$ /usr/bin/openssl version
OpenSSL 0.9.8zh 14 Jan 2016

so it looks like macOS still has an up to date openssl, why not use this version?

--------

Also, OSX has its own python version which comes compiled against 0.9.8

```
$ /usr/bin/python
Python 2.7.10 (default, Jul 30 2016, 18:31:42) 
>>> ssl.__file__
'/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ssl.pyc'
>>> ssl.OPENSSL_VERSION_INFO
(0, 9, 8, 29, 15)
>>> 
```

why not use the same openssl lib? why we need homebrew?

---------

> Scratch that about the headers not coming on the source, it seems at one point I looked at the source in the include folder which is empty, they come into ssl/ folder.

If this is no longer relevant, please update the PR description.

----------

Thanks for the work on this, but I feel that it needs more work and use the OpenSSL provided by the OS as this is the OpenSSL which is updated by Apple in case of issues.

If customers want an homebrew-based version, we can do that, but the default version should be the one based on the OpenSSL version provided by Apple


needs-changes

adiroiban:
@dumol lucky no 13... the build is green on macos 12.02 with the homebrew OpenSSL.

my problem is that homebrew feels like an hobby project and I could not find if there is a easy way to install 1.0.2 and not the latest openssl

for now homebrew might provide 1.0.2 but maybe in 1 month it will install 1.0.3 and then our version of python will break....

but... let's release a version and we will see what customers want 

I tried to add a check in install_dependencies to document that homebrew is required

@dumol please see if you can have this branch green on all builders and merge it.

Thanks!

adiroiban:
As discussed with @dumol we might want to rename this to osx-1012 to keep with the previous osx naming.

now... the problem is that this is not osx-1012 vanilla, but osx-1012-homebrew-openssl-1.0.2 ... and with homebrew I don't know how to pin a version.

but again, we can go with osx-1012 or macos-1012 and see how it it perceived by the customers.

my feeling is that this should be osx-hbopenssl102

dumol:
Not really about this Homebrew stuff either, but here it goes… The branch passes all tests: https://chevah.com/buildbot/builders/python-package-gk-review/builds/20

**needs-review**

src/python/Python-2.7.10/setup.py
```diff
@@ -831,9 +834,12 @@ def detect_modules(self):
                                          [ '/usr/sfw/lib' ] )
         else:
             ssl_libs = find_library_file(self.compiler, 'ssl',lib_dirs,
-                                     ['/usr/local/ssl/lib',
-                                      '/usr/contrib/ssl/lib/'
-                                     ] )
+                [
+                    # OpenSSL libraries for macOS are used from Homebrew.
```
adiroiban: I think that we should also mention here that order is important... and maybe this should be an inline comment to make it clear that it only refers to the line below... and not the the other files

src/python/Python-2.7.10/setup.py
```diff
@@ -805,6 +805,9 @@ def detect_modules(self):
                                libraries=math_libs) )
         # Detect SSL support for the socket module (via _ssl)
         search_for_ssl_incs_in = [
+                              # On MacOS OpenSSL headers are installed using
+                              # Homebrew. Make sure they are picked first.
+                              '/usr/local/opt/openssl/include',
                               '/usr/local/ssl/include',
                               '/usr/sfw/include/',
```
adiroiban: I know that this was not done in this branch, but maybe we should add a comment for each line from here to describe why we got it here in the first place.. so that we don't have to search the history

chevah_build
```diff
@@ -203,6 +204,16 @@ case $OS in
         export CFLAGS="$CFLAGS -mmacosx-version-min=10.8"
         export MACOSX_DEPLOYMENT_TARGET=10.8
         ;;
+    macos*)
+        # On macOS 10.12 or newer we need the Homebrew version of OpenSSL,
+        # as Apple only provides an old OpenSSL version: 0.9.8.
+        export CC="clang"
+        export CXX="clang++"
```
adiroiban: do we need this? AFAIK our code should not touch any C++ files

adiroiban:
Looks good. Happy to see it green.

@dumol I did not see any comment related to your initial suggestion of having this as osx1012.

Also, in the previous description of the PR you were listed as the requested reviewer. I have updated it so that I am a reviewer. 
Please check that the current PR description is valid ... and also the inline comments

Thanks!

needs-changes

dumol:
Actually, I just edited my last comment to correct a grammar slip. Oops…

**needs-changes**

src/python/Python-2.7.10/setup.py
dumol: Sure… I know `/usr/sfw/include` is needed in Solaris 10. And the other two I have just verified, we have inherited them from upstream Python.

chevah_build
dumol: It's true `$CXX` is not really needed, but it helps to define it, especially when debugging the build. However, `$CPPFLAGS` is needed. This is documented above when `CXX` is defined for the first time:
```
# CXX is not really needed, we export it to make sure g++ won't get picked up
# when not using gcc, and thus silence the associated configure warning.
# However, we set $CPPFLAGS later for linking to statically-compiled libs.
export CXX='g++'
```

src/python/Python-2.7.10/setup.py
dumol: Will do.

dumol:
Thank you for the review, have implemented most of the suggestions. Please check the commit above and the related comments.

In regards to the `macos-1202` name, I would normally be strongly against using `macos` instead of `osx` just because of the marketing shift from Apple. But seeing how this is actually quite different from the usual builds (because it needs the Homebrew's OpenSSL), I agree to differentiate it by using a different moniker for the OS.

**needs-review**

adiroiban:
thanks. changes-approved

dumol:
I wondered why the new `/usr/local` dependencies were not detected by the test script… There's a test there for pre-approved deps  and there was nothing added by this branch to that hard-coded list. 

After a bit of investigation, have discovered that the deps check in the test phase only works for OS'es where we check for `readline`, as the code was moved to a dedicated function and the call for the new function is indented too much. More at https://github.com/chevah/python-package/commit/d502092fb883bbb150f6eef145e9e752499ccf94#diff-d928fcacc543beb698c192bc0a02e064L349

If I fix the indentation, the following is output'ed for the test phase:
```
bs1f-macos-1012-x64-54:build buildslave$ ./python2.7-macos1012-x64/bin/python test_python_binary_dist.py
python 2.7.10 (default, Jan 19 2017, 02:25:40)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)]
zlib 1.0
stdlib ssl OpenSSL 1.0.2j  26 Sep 2016
pyopenssl 16.0.0 - OpenSSL OpenSSL 1.0.2j  26 Sep 2016
PyCrypto 2.6.1
ctypes 1.1.0
setproctitle 1.1.10
Got unwanted deps:
        /System/Library/Frameworks/Security.framework/Versions/A/Security
        /usr/lib/libgcc_s.1.dylib
        /usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib
        /usr/local/opt/openssl/lib/libssl.1.0.0.dylib
```
needs-changes

dumol:
Have fixed the dependencies check in master, merged the changes and adapted the check for the osx/macos dual nature of our support for this OS. All green `test_review` results: https://chevah.com/buildbot/builders/python-package-gk-review/builds/23

**needs-review**

adiroiban:
thanks.

changes-approved

PR #66, opened 2017-01-20
[#3810] Fix the test for checking dependencies.

dumol:
Scope
=====

The deps check in the test phase only works for OS'es where we check for `readline`, as the code was moved to a dedicated function and the call for the new function is indented too much. More at ​https://github.com/chevah/python-package/commit/d502092fb883bbb150f6eef145e9e752499ccf94#diff-d928fcacc543beb698c192bc0a02e064L349

Changes
=======

Unindented the call to the function that checks the deps, so that it is always called.
Added sanctioned deps for FreeBSD.
Tuned the deps for Linux in accordance to latest developments.

Drive-by test:
  * improved the deps check in OpenBSD to account for any unusual deps not in `/usr`, eg. `/opt` or `/home`.

reviewers: @adiroiban 

Please review the changes.
Run the tests.

### DIFF
--- a/python-modules/chevah-python-test/test_python_binary_dist.py
+++ b/python-modules/chevah-python-test/test_python_binary_dist.py
@@ -89,7 +89,7 @@ def get_allowed_deps():
                 'libtinfo.so.5',
                 ])
         else:
-            # Debian 7 x64 aka linux-x64 needs this for some reason.
+            # Debian 7 x64 (aka linux-x64) needs this for cffi.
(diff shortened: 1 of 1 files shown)


dumol:
`test_review` passed with deps checking for all OS'es in the _stable_ target for python-package: https://chevah.com/buildbot/builders/python-package-gk-review/builds/21

**needs-review**

adiroiban:
looks good. thanks!

changes-approved

PR #67, opened 2017-01-26
[#3827] Clean temporary directory on OSX.

brunogola:
Scope
=====

On OSX the temporary pip directory was not removed if pip failed as it is not under /tmp.

Changes
=======

Add code to clean the folder under ${TMPDIR}.

reviewers: @adiroiban @dumol 

Please check if changes make sense.

### DIFF
--- a/paver.sh
+++ b/paver.sh
@@ -98,12 +98,12 @@ clean_build() {
     fi
     # In some case pip hangs with a build folder in temp and
     # will not continue until it is manually removed.
-    rm -rf /tmp/pip*
-
     # On the OSX build server tmp is in $TMPDIR
-    # check if TMPDIR is set before trying to clean it.
     if [ ! -z "${TMPDIR-}" ]; then
+        # check if TMPDIR is set before trying to clean it.
         rm -rf ${TMPDIR}/pip*
+    else
+        rm -rf /tmp/pip*
     fi
 }
 
--- a/paver.sh
+++ b/paver.sh
@@ -99,6 +99,12 @@ clean_build() {
     # In some case pip hangs with a build folder in temp and
     # will not continue until it is manually removed.
     rm -rf /tmp/pip*
+
+    # On the OSX build server tmp is in $TMPDIR
+    # check if TMPDIR is set before trying to clean it.
+    if [ ! -z "${TMPDIR-}" ]; then
+        rm -rf ${TMPDIR}/pip*
+    fi
 }
 
 


brunogola:
needs-review

brunogola:
needs-review

adiroiban:
don't forget the ticket id in the title.

needs-review

paver.sh
```diff
@@ -99,6 +99,12 @@ clean_build() {
     # In some case pip hangs with a build folder in temp and
     # will not continue until it is manually removed.
     rm -rf /tmp/pip*
+
+    # On the OSX build server tmp is in $TMPDIR
+    # check if TMPDIR is set before trying to clean it.
+    if [ ! -z "${TMPDIR-}" ]; then
+        rm -rf ${TMPDIR}/pip*
+    fi
```
adiroiban: maybe and an else here and put the above ` rm -rf /tmp/pip*` ... so that when we have TMPDIR we only delete TMPDIR and not other folders

paver.sh
```diff
@@ -99,6 +99,12 @@ clean_build() {
     # In some case pip hangs with a build folder in temp and
     # will not continue until it is manually removed.
     rm -rf /tmp/pip*
+
+    # On the OSX build server tmp is in $TMPDIR
```
adiroiban: this comment  should be inside the block which deals with the osx condition

adiroiban:
looks good. ... only a few minor comments

changes-approved

paver.sh
```diff
@@ -98,7 +98,13 @@ clean_build() {
     fi
     # In some case pip hangs with a build folder in temp and
     # will not continue until it is manually removed.
-    rm -rf /tmp/pip*
+    # On the OSX build server tmp is in $TMPDIR
+    if [ ! -z "${TMPDIR-}" ]; then
```
dumol: Two minor questions/suggestions:
  * why use `! -z` instead of `-n`?
  * why use parameter expansion when we are not interested in corner cases here?

brunogola: `-n`  works. 

the parameter expansion is used because if we test only for `${TMPDIR}` we might get 'unbound variable' error when it is not defined.

thanks!

dumol: Not sure what you mean… I tried the `bash` in the AIX 5.3 build slave, the oldest we've got:
```
-bash-3.2$ bash --version
GNU bash, version 3.2.16(1)-release (powerpc-ibm-aix5.2.0.0)
Copyright (C) 2005 Free Software Foundation, Inc.
-bash-3.2$ if [ -n "$NOSUCHDIR" ]; then echo 'we have $NOSUCHDIR!'; fi
-bash-3.2$ export NOSUCHDIR="/tmp"
-bash-3.2$ if [ -n "$NOSUCHDIR" ]; then echo 'we have $NOSUCHDIR!'; fi
we have $NOSUCHDIR!
```
I think using parameter expansion here is overkill and will confuse us later.

brunogola: actually for some reason `-n` would give us a false positive:

```
[bgola@charango noisegrid]$ echo ${xx}
bash: xx: unbound variable
[bgola@charango noisegrid]$ if [ ! -z ${xx-} ]; then echo "not set"; fi
[bgola@charango noisegrid]$ if [ -n  ${xx-} ]; then echo "not set"; fi
S
```

so I will keep `! -z`

dumol: Because you use two tricks and they work together. But the easiest thing would be to use no trick at all…

brunogola: our script runs with `set -u`, so:

```
$ if [ -n "$NOSUCHDIR" ]; then echo 'we have $NOSUCHDIR!'; fi
bash: NOSUCHDIR: unbound variable
```

dumol: Ah, true, I forgot about the draconian settings of `paver.sh`. This is because of `set -o nounset`. Parameter extensions is indeed the solution for that, I have also used it at https://github.com/chevah/python-package/blob/master/python-modules/chevah-python-test/get_binaries_deps.sh#L19

Interestingly the `-n` and `-z` don't work equally here… I looked at the manual and couldn't figure it out. But is seems as though `+` would be better than `-` here, from what I gather at https://www.gnu.org/software/bash/manual/bashref.html#Shell-Parameter-Expansion, as `+` doesn't substitute anything if the parameter is null or unset. However, in my tests, both work for this check in the bash shell included with OS X.

PR #68, opened 2017-02-08
[#3855] Custom Python versioning on Linux/Unix.

dumol:
Scope
=====

The `python` binary in `python-package` should advertise a custom version through `sys.version`.
The custom version should be based on the current git revision.
The python archive name should include this custom version, eg. `python-2.7.10.c3dea23-linux-x86.tar.gz`.

Changes
=======
Added a `sed` command in python's `chevahbs` script to hijack an unused Mercurial-related string which would have always been `default` downstream of us. This is how the result looks like (the short form of the git revision is displayed before the build date):

```
Python 2.7.10 (ba67e06, Feb 13 2017, 14:08:38) 
[GCC 4.8.4] on linux2
>>> import sys
>>> sys.version
>>> '2.7.10 (ba67e06, Feb 13 2017, 14:08:38) \n[GCC 4.8.4]'
```
Added a python test for checking the presence of the short form of the git revision in python's `sys.version` string. When failing, the check errors out as follows:
```
% ./chevah_build test
~/dumol/python-package/build ~/dumol/python-package
python 2.7.10 (18aebdb9, Feb 13 2017, 18:38:53) 
[GCC 4.8.4]
zlib 1.0
stdlib ssl OpenSSL 1.0.1f 6 Jan 2014
cryptography 1.3.1 - OpenSSL OpenSSL 1.0.1f 6 Jan 2014
pyopenssl 16.0.0 - OpenSSL OpenSSL 1.0.1f 6 Jan 2014
PyCrypto 2.6.1
ctypes 1.1.0
setproctitle 1.1.10
Python's version doesn't match git rev!
        Bin ver: 18aebdb
        Git rev: a896862
PWD : /home/chevah/dumol/python-package/build
Fail: ./python2.7-ubuntu1404-x64/bin/python test_python_binary_dist.py

```
Added the git rev string in the python package archive and stopped using time stamps.

**Drive-by changes**:
  * removed duplicate python tests
  * cleaned up the error code scheme even more.

reviewers: @adiroiban

Please review the changes.
Build and test `python-package`, eg. https://chevah.com/buildbot/builders/python-package-gk-review/builds/29
Check `sys.version` for the `python` binary.
Check the file name for the newly created archive.
Check that the test for the git rev works (eg. by testing with a different python binary).

### DIFF
--- a/functions.sh
+++ b/functions.sh
@@ -180,7 +180,7 @@ make_dist(){
     target_tar_gz=${target_tar}.gz
 
     tar_gz_file=${target_folder}.tar.gz
-    tar_gz_timestamp_file=${target_folder}-${TIMESTAMP}.tar.gz
+    tar_gz_timestamp_file=${target_folder}-${TIMESTAMP}-${PYTHON_PACKAGE_VERSION}.tar.gz
 
     # Create a clean dist folder.
     execute rm -rf ${DIST_FOLDER}
--- a/functions.sh
+++ b/functions.sh
@@ -176,8 +176,8 @@ make_dist(){
     kind=$1
     target_folder=$2
 
-    target_tar=../dist/${kind}/${OS}/${ARCH}/${target_folder}-${TIMESTAMP}.tar
-    target_tar_gz=../dist/${kind}/${OS}/${ARCH}/${target_folder}-${TIMESTAMP}.tar.gz
+    target_tar=../dist/${kind}/${OS}/${ARCH}/${target_folder}-${TIMESTAMP}-${PYTHON_PACKAGE_VERSION}.tar
+    target_tar_gz=${target_tar}.gz
 
     tar_gz_file=${target_folder}.tar.gz
     tar_gz_timestamp_file=${target_folder}-${TIMESTAMP}.tar.gz
--- a/functions.sh
+++ b/functions.sh
@@ -182,7 +182,7 @@ make_dist(){
     target_tar_gz=${target_tar}.gz
 
     tar_gz_file=${target_folder}.tar.gz
-    tar_gz_target_file=${target_common}.tar.gz
+    tar_gz_source_file=${target_common}.tar.gz
 
     # Create a clean dist folder.
     execute rm -rf ${DIST_FOLDER}
@@ -197,7 +197,7 @@ make_dist(){
 
     # Create symlink.
     execute pushd ${DIST_FOLDER}/${kind}
-        execute ln -sf ${OS}/${ARCH}/${tar_gz_target_file} ${tar_gz_file}
+        execute ln -sf ${OS}/${ARCH}/${tar_gz_source_file} ${tar_gz_file}
     execute popd
 }
 
--- a/functions.sh
+++ b/functions.sh
@@ -177,11 +177,12 @@ make_dist(){
     target_folder=$2
 
     target_path=../dist/${kind}/${OS}/${ARCH}
-    target_tar=${target_path}/${target_folder}-${TIMESTAMP}-${PYTHON_PACKAGE_VERSION}.tar
+    target_common=python-${PYTHON_BUILD_VERSION}.${PYTHON_PACKAGE_VERSION}-${OS}-${ARCH}
+    target_tar=${target_path}/${target_common}.tar
     target_tar_gz=${target_tar}.gz
 
     tar_gz_file=${target_folder}.tar.gz
-    tar_gz_timestamp_file=${target_folder}-${TIMESTAMP}-${PYTHON_PACKAGE_VERSION}.tar.gz
+    tar_gz_target_file=${target_common}.tar.gz
 
     # Create a clean dist folder.
     execute rm -rf ${DIST_FOLDER}
@@ -196,7 +197,7 @@ make_dist(){
 
     # Create symlink.
     execute pushd ${DIST_FOLDER}/${kind}
-        execute ln -sf ${OS}/${ARCH}/${tar_gz_timestamp_file} ${tar_gz_file}
+        execute ln -sf ${OS}/${ARCH}/${tar_gz_target_file} ${tar_gz_file}
     execute popd
 }
 
--- a/functions.sh
+++ b/functions.sh
@@ -176,7 +176,8 @@ make_dist(){
     kind=$1
     target_folder=$2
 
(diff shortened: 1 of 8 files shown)


dumol:
This is a proof of concept, feedback welcome…

**needs-review**

brunogola:
I'm not sure what is the goal with this custom version, but if we are using the date, wouldn't it be better if we export that from the build script automatically so we don't need to change the serial value every time we build? 

regarding the format, i think that's fine.

adiroiban:
I think that Misu's work is proof of concept.
Yes. We should export the version from the build script... maybe via a env var.

I would like to keep this change as small as posible so just inject the version in PY_VERSION.

---------

The goal here is to be able to identify the exact python-package release which was used to run a test and to create a distributable package.

------------

Right now we have the Arch linux test passing on chevah/server ... and the local `paver dist` working, yet the staging version is failing.

We have versions for each software used by our `paver deps`  command, with the exception of the python distribution itself.

---------

The goal is to create an environment in which we can detect failures similar to ArchLinux ... or at least to help us investigate that.

I think that keeping a note in python version is one step

adiroiban:
so... for me is ok even if we have

```
$ build/Python-2.7.10/python                                  
Python 2.7.10 (default, Feb  8 2017, 20:12:45, a3da34ed3) 
```

I would not add a separate release so that we will not trigger some side effects.... I don't know what software might use the version for doing funny things.

Thanks!

needs-changes

dumol:
Thanks for the suggestions, have tried a different approach, no C code changes this time, just a `sed` command to hijack an unused Mercurial-related string which would have always been `default` downstream of us. This is how the result looks like (the first 8 letters of the git revision are displayed before the build date):

```
Python 2.7.10 (863ba9ad, Feb 11 2017, 13:39:45)
[GCC 4.8.4] on linux2
>>> import sys
>>> sys.version
>>> '2.7.10 (863ba9ad, Feb 11 2017, 13:39:45) \n[GCC 4.8.4]'
```
What do you think?

**needs-review**

dumol:
As I have reversed the previous changes in the same commit, I recommend taking a look at the changes as a whole. Thanks!

adiroiban:
thanks. looks pretty good.

Since we can't use the sha for version comparison I think that the current solution is excellent :)

----------

This can be done in a separate ticket... but maybe is a easy fix.

To help identify a version, maybe we can include this 8 chars in the python archive name... beside the date.... or instead of the date :)

changes-approved

dumol:
Thank you for the feedback, I'm working on getting this into shape, thus:

**needs-changes**

dumol:
I think I've implemented all suggestions… Not sure about the `paver.sh` thing though, do you want to download through it a specific version of the python-package archive (eg. `python2.7-ubuntu1404-x64-20170213-d2361052.tar.gz`) instead of the generic one (eg. `python2.7-ubuntu1404-x64.tar.gz`), as we currently do? For now we depend on symbolic links on the server hosting the packages and the long name is not that relevant.

**needs-review**

functions.sh
```diff
@@ -180,7 +180,7 @@ make_dist(){
     target_tar_gz=${target_tar}.gz
 
     tar_gz_file=${target_folder}.tar.gz
-    tar_gz_timestamp_file=${target_folder}-${TIMESTAMP}.tar.gz
+    tar_gz_timestamp_file=${target_folder}-${TIMESTAMP}-${PYTHON_PACKAGE_VERSION}.tar.gz
```
adiroiban: I think that we should remove the timestamp .... as the timestamp is arbitrary and does not help with versioning. Thanks!

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -483,6 +459,20 @@ def main():
...
+    except:
+        sys.stderr.write('Could not get the git rev for the current tree.\n')
+        exit_code = 17
+    else:
+        bin_ver = sys.version.split('(')[1][:7]
+        if git_rev != bin_ver:
+            print "Python's sys.version doesn't seem to match the current git rev..."
+            print "\tGit rev:" , git_rev
+            print "\tBin ver:" , bin_ver
+            exit_code = 18
```
dumol: I would appreciate if someone enlightens me how come the `\t` in the lines above doesn't indent with a tab, as it does in `test_dependencies()`.

adiroiban:
I would like to get rid of symbolic links and grab the python based on the OS name + arch + python version + sha.

I would like the python archive to include the exact python version like '2.7.10' 

so the archive name would be

```
python-2.7.10.c3dea23-linux-x86.tar.gz
```

in paver.conf instead of `PYTHON_VERSION='python2.7'` we can have something like

```
PYTHON_VERSION='default@2.7.11.23d24:solaris10-x64@2.7.8.53da3f4:sles10-x64@2.7.11.578ac4f2'
```

then paver.sh will decide which version to install based on PYTHON_VERSION....

and in an ideal word, each time paver.sh is called it will also look at the existing python_version and reinstall it if the existing version is different to the one from paver.conf

adiroiban:
Changes looks good.

You can leave or remove the timestamp from the archive name.

For now we will still use the symbolic link as we also need to update the Windows version.

After that we can look at implementing PYTHON_VERSION changes in a separate ticket.

Thanks!

changes-approved

brunogola:
changes-approved

dumol:
Have implemented the suggestions from the last review and updated the PR. @adiroiban please take a look at the results and let me know if they are fine. (Have removed @bgola from list of reviews for this). Thanks!

**needs-review**

adiroiban:
i like it a lot :) great work.

just a small change request.
please also keep the version in a file.. maybe lib/PYTHON_VERSION or lib/python-release 

and lib is found in both windows and unix version.

looking forward to see this hitting production.

changes-approved

adiroiban:
> Not sure about the paver.sh thing though, do you want to download through it a specific version of the python-package archive (eg. python2.7-ubuntu1404-x64-20170213-d2361052.tar.gz) instead of the generic one (eg. python2.7-ubuntu1404-x64.tar.gz), as we currently do? For now we depend on symbolic links on the server hosting the packages and the long name is not that relevant.

the paver.sh will be done in a separate ticket.

for backward compatibility we will still need to keep the symlink.

the plan is for paver.sh to download directly the archive and re-download it ... in this way we can have different branches with different python versions

thanks!

PR #72, opened 2017-03-30
[#3966] Support for OpenSSL 1.0.x in HP-UX.

dumol:
Scope
=====

As noted in Trac's ticket #2703, the HP-UX initial support was done for OpenSSL 0.9.8, but only 1.0.2 is currently supported in HP-UX, according to https://h20392.www2.hpe.com/portal/swdepot/displayProductInfo.do?productNumber=OPENSSL11I.

Changes
=======

Added an explicit import for `_ssl` to avoid errors with `pip` and `cffi` when building against OpenSSL 1.0.2.
Do not build CFFI on HP-UX for now.
Updated OpenSSL deps to 1.0.x for the test phase.

reviewers: @adiroiban 

Please review the changes.
Run the tests, eg. https://chevah.com/buildbot/builders/python-package-gk-review/builds/44

### DIFF
--- a/python-modules/cffi-1.5.2/setup.py
+++ b/python-modules/cffi-1.5.2/setup.py
@@ -1,6 +1,7 @@
 import sys, os
 import subprocess
 import errno
+import _ssl
 
 
 sources = ['c/_cffi_backend.c']
--- a/python-modules/cffi-1.5.2/setup.py
+++ b/python-modules/cffi-1.5.2/setup.py
@@ -1,7 +1,7 @@
 import sys, os
 import subprocess
 import errno
-import _ssl
+import _ssl # Needed in HP-UX, unexpectedly.
 
 
 sources = ['c/_cffi_backend.c']
--- a/chevah_build
+++ b/chevah_build
@@ -218,6 +218,9 @@ case $OS in
         export MAKE="gmake"
         export BUILD_ZLIB="yes"
         export BUILD_LIBEDIT="no"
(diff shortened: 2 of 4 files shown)


python-modules/cffi-1.5.2/setup.py
```diff
@@ -1,6 +1,7 @@
 import sys, os
 import subprocess
 import errno
+import _ssl
```
adiroiban: I think that we need a comment to describe why we are doing this awkward import.

but with the current build. do we need it?

dumol: We don't need this particular one, but we'll need it if/when we'll get back to building CFFI on HP-UX. I judged it harmless, but I guess I should add comments for these extra imports.

dumol:
**needs-review**

adiroiban:
thanks. looks good.

The fact that _ssl import is needes is a bug with the Python dynamic module loading.

We have observed the same bug in RHEL4.

Instead of importing the private _ssl, we went for importing the high level public hashlib ... which will trigger the _ssl import

I think that this is fine.

Right now I don't have time to work more in this python-package.

In order to move forward with the HP-UX issue so that we can use the latest OpenSSL, we should first integrate the automated tests from python-cryptography and pyopenssl 

in a similar way in which we are executing the tests for scandir

thanks!

changes-approved

dumol:
Wrong copy/paste in the commit message. :-/

PR #75, opened 2017-04-26
[bad-id] Use a fixed version of pip and setuptools

adiroiban:
see #77


adiroiban:
needs-review

adiroiban:
as a side now, I saw that the src/python/chevahbs is using both `patch` and `patch -p0` ... is this on purpose?

adiroiban:
I think that the current version is patching the source get-pip.py ... but we should have a get-pip.py in build and patch and execute that version.

Will update this soon... but you can still review it

thanks

PR #76, opened 2017-04-26
[#4020] Patch the HP-UX build to load the right libraries.

adiroiban:
Scope
=====

The latest 2.7.13 build is failing on HPUX

Why we got this
=============

* No complete tests for the new distribution when used with our full build and test system

Changes
=======

Add a patch for site.py on HPUX

The tests were extended to run a full test for compat.

As a drive-by change I have documented the basic principles of generating a patch.

reviewers: @dumol 

this depends on #77 . I have include it here as otherwise the build is blocked

check that changes make sense.

for now it is using a specific branch from chevah/compat.... but I hope that branch will be merged soon into master and then I will update the code to use master

thanks


adiroiban:
for now the ./chevah_build test_compat is a separate step

one test is failing... but it does not seem like a critical error in python-package

I have updated the buildbot to run test_compat in separate step so that it can fail... I will soon update the buildbot configuration to  mark it just as a warning

I think that we can merge it as it is and then have a separate issue if there are more errors.

needs-review

src/python/chevahbs
```diff
@@ -31,6 +36,8 @@ chevahbs_patch() {
             # Our patch to use -fPIC extra argument for linking only with GCC.
             echo "Applying hpux_fpic.patch:"
             execute patch < hpux_fpic.patch
+            echo "Applying hpux_site.patch:"
+            execute patch -p0 < hpux_site.patch
```
adiroiban: True. Thanks!

README
```diff
@@ -32,3 +32,24 @@ Do a production build
 (files go in http://binary.chevah.com/production/python/)::
 
     paver test_remote NAME_OF_BUILDER --force-upload-production
+
+
+Patching upstream code
+----------------------
+
+To simplify changes management we are keeping the changed we done from upstream
+as separate patch files and apply them at build time.
```
adiroiban: much better :) thanks!

chevah_build
```diff
@@ -685,6 +698,41 @@ command_test() {
...
+    # the local version.
+    echo '##### Running chevah.compat tests... #####'
+    execute rm -rf compat
+    execute git clone https://github.com/chevah/compat.git --depth=1 -b master
+    execute pushd compat
+    # We patch the Python version to match the one that we have just built
+    # and copy it in cache so that it will be picked up by the new build.
+    # Make sure we have an empty line at the end.
+    execute echo '' >> paver.conf
+    execute echo "PYTHON_CONFIGURATION=${PYTHON_BUILD_VERSION}.${PYTHON_PACKAGE_VERSION}" >> paver.conf
```
adiroiban: thanks

README
```diff
@@ -32,3 +32,24 @@ Do a production build
...
+With that in mind we can have multiple patches for the same OS.
+
+Here is an example for creating a patch for
+src/python/Python-2.7.13/Lib/site.py::
+
+    cd src/python/Python-2.7.13/
+    cp Lib/site-hpux.py Lib/site-patched.py
+    # Change Lib/site-patched.py as you want then create the diff.
+    diff -ru Lib/site.py Lib/site-patched.py
+    # Copy the diff result, including the command line into
```
adiroiban: into the patch file at.

yes. it was poorly written. thanks!

README
```diff
@@ -32,3 +32,24 @@ Do a production build
...
+
+To keep things as simple as possible, each path should target on a specific
+change.
+With that in mind we can have multiple patches for the same OS.
+
+Here is an example for creating a patch for
+src/python/Python-2.7.13/Lib/site.py::
+
+    cd src/python/Python-2.7.13/
+    cp Lib/site-hpux.py Lib/site-patched.py
```
adiroiban: true.

adiroiban:
> However, some exotic build slaves might not pass compat tests, in which case we might move them from the all group to unstable, if we want to rely on this. Or possibly define a review group for python-package?

For now, the compat tests are just "informative".
If the test fails, the step will still succeed, as the step is not called with `execute`.

I am updating the compat build system to be faster...and to also pass on hpux. Hope that will reach compat master soon

In the future, I want to update buildmaster to allow "warning" state. That is a step which somehow fails, but the whole build is still considered successful. 

once we have Solaris 10 Sparc up and running, can can start adding a PQM for this project... and only leave out Solaria 11 Sparc.

Thanks for the great review!

PR #77, opened 2017-04-26
[#4024] Use a fixed version of pip and setuptools

adiroiban:
Scope
=====

This updates the build script for python to use a pinned version for pip and setuptools.

In previous code the latest version was used so if the pypi server was update, we would end up with different python distribution, for the same commit

Changes
=======

There is no way for get-pip.py to install a specific pip. it will always install the latest...so I had to patch it.

for setuptools there was a workarround 

reviewers: @dumol 

check that changes make sense

we have newer setuptools in our pypi,chevah.com mirror.

you can try updating the chevah_build with an even older pip version and check that the requiest version is used

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -700,6 +700,18 @@ command_test() {
 }
 
 
+#
+# Test the newly created Python binary dist.
+#
+help_text_test_compat=\
+"Run a full test for the chevah/compat master."
+command_test_compat() {
+    # For now it does nothing.
+    # Here since the buildmaster was updated.
+    # Will be overwritten in #4020.
+}
+
+
 help_text_publish_python=\
 "Upload Python binaries for current OS."
 command_publish_python() {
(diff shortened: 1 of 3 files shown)


adiroiban:
needs-review

chevah_build
```diff
@@ -19,6 +19,12 @@ LIBEDIT_VERSION="20160903-3.1"
 GMP_VERSION="6.1.2"
 SQLITE_VERSION="3.17.0"
 PYSQLITE_VERSION="2.6.3"
+# These are used by the patched get-pip.py.
+# PIP_VERSION is already used by get-pip for something else.
+# For now, we can't use version 8 of pip as our PyPi server does not support
+# simplified names.
+export CHEVAH_PIP_VERSION="7.1.2"
```
dumol: For uniformity, there's a line further down for exporting needed vars. It is prepended by _# Export the variables needed by the chevahbs scripts and the test phase._

chevah_build
```diff
@@ -540,7 +546,12 @@ command_build_sqlite() {
 command_build_python_extra_libraries() {
 
     # Install the latest PIP and setuptools.
-    execute $PYTHON_BIN python-modules/get-pip.py $PIP_ARGS
+    # But first path it to get the exact version.
+    # get-pip will alwasy try to get the latest version, so we ask it to not
```
dumol: _alwasy_ -> _always_

PR #79, opened 2017-05-11
[#4040] No more generic symlinks.

dumol:
Scope
=====

we have start pushing the new build system in all project ... and those repos which were not yet updated, should be updated.

for this we should stop creating symlinks in the python package as they are no longer used.

With this, we should update the paver.sh with the latest version from brink ... to keep them up to date

Changes
=======

Have removed the code for generating symbolic links per OS/ARCH.
Have updated `paver.sh` from `brink` with related changes in `paver.conf` and `pavement.py` (updated used brink/twisted versions to latest).

**Drive-by change**:
  * got rid of the count to 16 while setting version per platform (code manually tested on the exotic platforms)
  * updated `get-pip.py` to latest version and set pip to 9.0.1.

reviewers: @adiroiban 

Please review the changes.
Run `./paver.sh deps`.
Run `./chevah_build build` and notice there is no generic symlink in `dist/` anymore.
Run the automated tests, eg. https://chevah.com/buildbot/builders/python-package-gk-review/builds/52


dumol:
**needs-review**

adiroiban:
thanks. changes-approved

dumol:
I can't come to terms with the count to 16, I think we can do better…

**needs-changes**

adiroiban:
:)

Thanks.

If it works is ok.

For updating the latest pip, the important part is to update the PIP version in the chevah_build.

updating get-pip.py is optional.

----------

We still the patched get-pip.py as otherwise it will install the latest pip available in ou pypi.chevah.com and will ignore what we have configured in our script.

So if we run it now it will install pip 9.0.1.... but if later we copy pip 12.3.0 in our pypi.chevah.com , the same old get-pip.py will install the latest (now 12.3.0) version,

We don't want that.

We want a build, which when re-build at any time in the future, will use the same versions.

For the python-package we want to always pin all our dependencies.

needs-changes

dumol:
OK, I see… I thought we want latest pip from now on.

dumol:
OK, I have redone the last commit. 

`get-pip.py` is latest version which accepts custom pip versions through a command switch, so no need for the patch and for exporting CHEVAH_PIP_VERSION.

`pip` is set to 9.0.1, but I have also tested 7.1.2 with this new arrangement, and it worked in Ubuntu 14.04 on Deag with the usual warning that pip is not the latest version.

Have tried to also update `setuptools` to latest version, but run into more problems and could figure them on the spot. Please give it a try if you have the time.

**needs-review**

chevah_build
```diff
@@ -21,9 +21,7 @@ SQLITE_VERSION="3.17.0"
 PYSQLITE_VERSION="2.6.3"
 # These are used by the patched get-pip.py.
 # PIP_VERSION is already used by get-pip for something else.
-# For now, we can't use version 8 of pip as our PyPi server does not support
-# simplified names.
-CHEVAH_PIP_VERSION="7.1.2"
+CHEVAH_PIP_VERSION="9.0.1"
```
adiroiban: we can named it PIP_VERSION ... as we will no longer export it

adiroiban:
thanks.

changes-approved

dumol:
https://chevah.com/buildbot/builders/python-package-gk-review/builds/53

PR #80, opened 2017-05-31
[#4132] Ubuntu 16.04 ARM64 support.

dumol:
Scope
=====

`python-package` builds fine on the new Pine64 board, but there's one additional dep on Ubuntu 16.04 compared to the arm64 build on Ubuntu 14.04. (Or perhaps last arm64 build was affected by the bug fixed in #66.)

Changes
=======

Simply added the missing dep for ARM64 builds on Ubuntu 16.04.

reviewers: @adiroiban 

Please check the change.
Run the automated tests, eg. https://chevah.com/buildbot/builders/python-package-ubuntu-1604-arm64/builds/3.

### DIFF
--- a/python-modules/chevah-python-test/test_python_binary_dist.py
+++ b/python-modules/chevah-python-test/test_python_binary_dist.py
@@ -76,6 +76,10 @@ def get_allowed_deps():
             allowed_deps.extend([
                 'libtinfo.so.5',
                 ])
+            if 'arm64' in chevah_arch:
+                allowed_deps.extend([
+                    'libgcc_s.so.1',
+                    ])
         elif ('raspbian' in chevah_os):
             allowed_deps.extend([
                 'libcofi_rpi.so',


dumol:
**needs-review**

adiroiban:
looks good.  python build is faster than rpi2

chevah/server tests are running in 14 minutes vs 17 minutes as on rpi2 ... so there is an improvement.

I was hopping that we can have them run in 10 minutes... but this is also ok.

-------

As a side note, I think that we can remove the ubuntu 12 (x86 and x64) builders.

---------

the compat tests are failing... maybe due to missing / failing pam module, but this need to be handled in a separate ticket:

https://chevah.com/buildbot/builders/python-package-ubuntu-1604-arm64/builds/3/steps/test_1/logs/stdio

-----------

thanks

changes-approved

adiroiban:
I have also tried rpi3 build... and got a similar dependcy failure https://chevah.com/buildbot/builders/python-package-raspbian-8/builds/0/steps/test/logs/stdio ... but maybe this can be done in a separate ticket

dumol:
There's a separate ticket for Raspbian 8 support, as we also need to not use Raspbian 7 binaries. When done with both, I'll bring over all the changes to brink.

PR #82, opened 2017-06-07
[#4154] OpenBSD 6.1 support.

dumol:
Scope
=====

OpenBSD 6.1 is the latest stable version.
Current code doesn't build on OpenBSD 6.1.

Changes
=======

Use `wxneeded` flag when linking the Python binary in OpenBSD. In 6.1 it's not enough to mark the current filesystem as `wxallowed` for binaries to be exempted from the W^X policy.

Updated `cryptography` to latest version.
This also required newer `cffi` and `asn1crypto`.
For some OS'es, latest `pyOpenSSL` was needed.
Still, some older OS'es didn't build `cryptography`, so the following actions were taken:
  * for OS X 10.8, the upstream wheel was uploaded (this is used in MacOS 10.12 too)
  * for RHEL 5 and SLES 11 we make an exception and still use the old set of PIP_LIBRARIES.

**Drive-by fix**:
  * get the number of processors in Windows too to silence a warning during the test phase.

reviewers: @adiroiban 

Please review the changes.
Run the automated tests, eg. https://chevah.com/buildbot/builders/python-package-gk-review/builds/0

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -463,6 +463,7 @@ command_build() {
     # as copying the includes (HP-UX's linker fails with -I when not using GCC).
     mkdir -p $INSTALL_FOLDER/{include,lib}
     export LDFLAGS="-L${INSTALL_FOLDER}/lib/ ${LDFLAGS}"
+    export PKG_CONFIG_PATH="${INSTALL_FOLDER}/lib/pkgconfig/:${PKG_CONFIG_PATH}"
 
     if [ "$BUILD_LIBFFI" = "yes" ]; then
         build 'libffi' "libffi-$LIBFFI_VERSION" ${PYTHON_BUILD_FOLDER}
--- a/chevah_build
+++ b/chevah_build
@@ -55,7 +55,7 @@ EXTRA_LIBRARIES_NO_CFFI="\
 PIP_LIBRARIES="\
     setproctitle==1.1.10 \
     cryptography==1.9 \
-    pyOpenSSL==16.0.0 \
+    pyOpenSSL==17.0.0 \
     scandir==1.5 \
     "
 # Pre-combiled libraries distributed as wheels for Windows.
--- a/chevah_build
+++ b/chevah_build
@@ -54,7 +54,7 @@ EXTRA_LIBRARIES_NO_CFFI="\
 # List of python modules installed using pip
 PIP_LIBRARIES="\
     setproctitle==1.1.10 \
-    cryptography==1.3.1 \
+    cryptography==1.9 \
     pyOpenSSL==16.0.0 \
     scandir==1.5 \
     "
@@ -254,11 +254,6 @@ case $OS in
         export CC="clang"
         export CXX="clang++"
         export BUILD_LIBEDIT="no"
-        ;;
-    openbsd*)
-        # cryptography fails to build with LibreSSL, use the packaged OpenSSL.
-        export LDFLAGS="-L/usr/local/eopenssl/lib $LDFLAGS"
-        export CPPFLAGS="$CPPFLAGS -I/usr/local/include/eopenssl"
         ;;
     archlinux)
         export BUILD_LIBEDIT="no"
--- a/chevah_build
+++ b/chevah_build
@@ -751,9 +751,8 @@ command_test_compat() {
     # download it.
     execute mkdir cache
     execute cp -r ../$LOCAL_PYTHON_BINARY_DIST cache/
-    # For now, this test is just informative, so even if we fail, we are OK.
     # Some tests might fail due to causes which are not related to python.
-    ./paver.sh test_ci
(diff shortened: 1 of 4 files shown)


adiroiban:
just a drive by comment.

happy to see an updated cryptography.

You can look at also updating pyopenssl to 17.0.0

-----------

if we care about python cryptography and libressl we can look at sponsoring a jenkins builder/slave for a linux with libressl and then openbsd with libressl .

In this way, if there is an issue, there is change that the developers might look into this issue   

there is no guarantee that upstream developers will work for free for libressl, but we can try

:)

dumol:
**needs-review**

chevah_build
```diff
@@ -264,7 +263,8 @@ case $OS in
         # and the result is not portable.
         export BUILD_LIBEDIT="no"
         ;;
-    sles10)
+    sles10|sles11|rhel5)
+        # Up-to-date cryptography not working on these...
```
adiroiban: I don't know what to say about sles 11 and rhel5.

this change will brink back ancient pyopenssl version.

I think that is better to add a separate condition for them and set PIP_LIBRARIES to the  pyOpenSSL==16.0.0 and cryptography==1.3.1

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -421,7 +421,7 @@ def main():
 
             if chevah_os == 'windows':
                 # Check OpenSSL version on windows.
-                expecting = u'OpenSSL 1.0.2g  1 Mar 2016'
+                expecting = u'OpenSSL 1.1.0f  25 May 2017'
```
adiroiban: Great change.

We will need a follow up ticket on Windows for chevah/server in which we will update paver.conf to the to be release python-package version

and then update the SFTPPlus documentation to inform that on Windows we have OpenSSL 1.1.0

dumol:
Thanks! Have reverted to using the old set of PIP_LIBRARIES for rhel5/sles11. Tests currently running at https://chevah.com/buildbot/builders/python-package-group-all/builds/4.

**needs-review**

python-modules/chevah-python-test/test_python_binary_dist.py
dumol: Are you sure having different major OpenSSL versions in `stdlib` and `cryptography` is not an issue? I looked it up a bit, and the `cryptography` guys thinks it's fine for Windows and MacOS (at https://cryptography.io/en/latest/installation/#static-wheels).

Anyway, this is how it looks now:
```
##### Executing Chevah tests... #####
python 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:42:59) [MSC v.1500 32 bit (Intel)]
zlib 1.0
stdlib ssl OpenSSL 1.0.2j 26 Sep 2016
cryptography 1.9 - OpenSSL OpenSSL 1.1.0f  25 May 2017
pyopenssl 17.0.0 - OpenSSL OpenSSL 1.1.0f  25 May 2017
PyCrypto 2.6.1
ctypes 1.1.0
```

dumol:
I thought failing the compat tests will break the build. Do we have failed compat tests for this branch?

adiroiban:
> I thought failing the compat tests will break the build. Do we have failed compat tests for this branch?

When I have introduced the compat tests in python-package, there was a chicken and egg problem.

compat tests were failing due to python-package changes.... but the compat tests should have been updated to work with latest python-package

see: https://chevah.com/buildbot/builders/python-package-ubuntu-1604-arm64/builds/5/steps/test_1/logs/stdio

one tests is failing... but the step is successful

python-modules/chevah-python-test/test_python_binary_dist.py
adiroiban: it should be fine if we have different versions.

we are using stdlib OpenSSL only for sha hash operations.

All TLS/SSL and serious crypto stuff is done via Python cryptography

--------

but I don't know if our server code can handle all OpenSSL 1.1.0 changes... in the case in which there are backward incompatible changes.

dumol:
OK, have enabled failed builds for failed compat tests. But now I see Windows fails as well: https://chevah.com/buildbot/builders/python-package-win-xp-dev/builds/11

Is this serious or the chicken and egg problem?

dumol:
OK, have enabled failed builds for failed compat tests. But now I see Windows fails as well: https://chevah.com/buildbot/builders/python-package-win-xp-dev/builds/11

Is this serious or the chicken and egg problem?

dumol:
@adiroiban, when you have some time please look at the failed compat test for windows referenced above. Thanks!

dumol:
We talked this face to face during the team-building and the Windows errors seem innocuous, so this should land on master.

PR #83, opened 2017-07-03
[#4198] Support for SLES 11 with Security Module.

dumol:
Scope
=====

Build with the OpenSSL 1.0.x from the Security Module repo on SLES 11. Therefore, use updated cryptography.

Changes
=======

Patched `paver.sh` to check for `libopenssl1` RPM in SUSE 11.
Defined new `chevah_os` for this SLES11/OpenSSL 1.0.x combo as `sles11sm`.
Patched `chevah_build` to install `libopenssl1-dev` if needed in SLES 11 with SM.
Patched `test_python_binary_dist.py` to account for the `sles11sm` moniker.

**Drive-by fix:**
  * use `cryptography` 1.8.3 on OS X, as our 10.8 is not supported in 1.9.0.

reviewers: @adiroiban 

Build the package in SLES 11 SM and checked for updated cryptography linked to the OpenSSL 1.0.x lib from the Security Module.
Build the package in SLES 11 without SM to check for possible breakage.
Run the automated test suite.

### DIFF
--- a/paver.sh
+++ b/paver.sh
@@ -544,8 +544,8 @@ detect_os() {
                     "$os_version_raw" os_version_chevah
                 OS="sles${os_version_chevah}"
                 if [ ${os_version_chevah} -eq 11 ]; then
-                    # Check for SLES 11 with OpenSSL 1.0 (aka Security Module).
-                    if [ -n $(rpm -q openssl1) ]; then
+                    # In 11.x check for OpenSSL 1.0.x (aka Security Module).
+                    if [ -x /usr/bin/openssl1 ]; then
                         OS="sles11sm"
                     fi
                 fi
--- a/paver.sh
+++ b/paver.sh
@@ -67,7 +67,7 @@ PYTHON_PLATFORM='unknown-os-and-arch'
 PYTHON_NAME='python2.7'
 BINARY_DIST_URI='https://binary.chevah.com/production'
 PIP_INDEX='http://pypi.chevah.com'
-PAVER_VERSION='1.2.1'
+BASE_REQUIREMENTS=''
 
 # Load repo specific configuration.
 source paver.conf
@@ -218,17 +218,8 @@ write_default_values() {
 # Install base package.
 #
 install_base_deps() {
-    local base_packages
-    base_packages="paver==$PAVER_VERSION"
-    if [ "$BRINK_VERSION" = "skip" ]; then
-        echo "Skipping brink installation."
-    else
-        base_packages="$base_packages chevah-brink==$BRINK_VERSION"
-    fi
-
-    echo "Installing $base_packages."
-
-    pip_install "$base_packages"
+    echo "Installing base requirements: $BASE_REQUIREMENTS."
+    pip_install "$BASE_REQUIREMENTS"
 }
 
 
--- a/paver.sh
+++ b/paver.sh
@@ -543,11 +543,9 @@ detect_os() {
                 check_os_version "SUSE Linux Enterprise Server" 10 \
                     "$os_version_raw" os_version_chevah
                 OS="sles${os_version_chevah}"
-                if [ ${os_version_chevah} -eq 11 ]; then
-                    # In 11.x check for OpenSSL 1.0.x (aka Security Module).
-                    if [ -x /usr/bin/openssl1 ]; then
-                        OS="sles11sm"
-                    fi
+                # On 11.x, check for OpenSSL 1.0.x (a.k.a. Security Module).
+                if [ ${os_version_chevah} -eq 11 -a -x /usr/bin/openssl1 ]; then
+                    OS="sles11sm"
                 fi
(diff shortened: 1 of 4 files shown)


paver.sh
```diff
@@ -552,6 +552,12 @@ detect_os() {
                 check_os_version "SUSE Linux Enterprise Server" 10 \
                     "$os_version_raw" os_version_chevah
                 OS="sles${os_version_chevah}"
+                if [ ${os_version_chevah} -eq 11 ]; then
+                    # Check for SLES 11 with OpenSSL 1.0 (aka Security Module).
+                    if [ -n $(rpm -q openssl1) ]; then
```
adiroiban: do we need to execute RPM here? can we just look for a file / path which we know that will exists with OpenSSL 1.0 ?

I don't want to have the paver.sh scrip to heavy and dependent on to many executables

dumol: We could check for `/usr/bin/openssl1`, but the RPM check is stronger. And the `rpm` binary is practically guaranteed to exist on a SLES 11 system. No strong feelings though…

Perhaps we should also emit a message in the console in SLES 11 if OpenSSL 1.0.x is not found?

adiroiban: if you think that PRM is ok, then we can keep it.

no need to send a warning or any message... if openssl 1.0.1 is installed, we should go with it... if is missing we should go with 0.9.8

dumol:
`win-xp` and `ubuntu1605-arm64` still fail compat tests. Otherwise this branch would pass `test_review`, albeit not consistently, because of mysterious issues with `hpux` (hw problems?) and race conditions in compat tests on a few slaves (`sles10` and `obsd61`). Details at https://chevah.com/buildbot/builders/python-package-gk-review/builds/6.

**needs-review**

adiroiban:
I have updated paver.sh with the version from brink in which the bootstrap is simplified by BASE_REQUIREMENTS 

-----------

Compat bug 4183 is not yet fixed... hope it will be fixed soon.

---------

Is still not clear how the openssl version is selected by the build system.

Is the python build system smart enough so that it will auto-detect the newest openssl version?

It would help to have some words, somewhere (maybe in the chevah build script), to talk about SLES 11 SM usage.

--------

The build looks good https://chevah.com/buildbot/builders/server-sles-11-sm/builds/0

I hope that I will fix the compat soon and will follow up on the server branch to add the SLES 11 SM version

thanks!

changes-approved

adiroiban:
See https://github.com/chevah/server/pull/1227

it looks like the arch and osx-108  are failing due to cryptogrphy

and Windows fails due to some permissions errors on binary.chevah.com https://chevah.com/buildbot/builders/server-win-dc-client/builds/174/steps/deps/logs/stdio

and the other builds fails due to newer cryptography, which has different default values

we need to look into this... I think that some of the failures are not due to the changes in this branch (but from previous branches) but we need to fix them

needs-changes

paver.sh
dumol: We use `rpm` in `chevah_build` too, so it's not a new requirement at all.

dumol:
The build system worked with OpenSSL 1.0.x out of the box, the only requirement is to have OpenSSL 1.0.x libs and headers installed. There could be a problem if headers for both versions of OpenSSL are installed, I haven't tested. And it's not our problem, as we control both build slaves.

paver.sh
adiroiban: well. paver.sh is used outside of the python-package so we are pushing this dependency to all places where paver.sh is used on sles

dumol:
The Windows builds fails because it cannot find a fallback package.
And there is no fallback package because the link is not updated any more in `python-package`.
It tried to get a fallback package because there is no Python package for the current testing version.
That one is missing because there was no successful build of current `python-package`.
And this happens because the `compat` tests fail consistently for Windows.

paver.sh
dumol: I'm aware of that. I'll change the check if you insist, no hard feelings..

dumol:
After manually fixing the symbolic link for the fallback version in testing, the build passes: https://chevah.com/buildbot/builders/server-win-dc-client/builds/184

dumol:
I've set OS X to use cryptography 1.8.2, the last version that supports 10.8. Please retry with latest revision.

As for Arch, as far as I can tell, it seems to have the same issues with the default values that are changed in latest cryptography that affects the others slaves. Please double-check. Thanks!

Tests associated with this rev:  https://chevah.com/buildbot/builders/python-package-gk-review/builds/9

**needs-review**

dumol:
`python-2.7.13.25cf4cf` packages from all build slaves have been uploaded in production in https://chevah.com/buildbot/builders/python-package-group-all/builds/7, https://chevah.com/buildbot/builders/python-package-hpux/builds/23 and https://chevah.com/buildbot/builders/python-package-raspbian-8/builds/19.

PR #84, opened 2017-07-21
[#4244] Build only 32bit binaries on Solaris.

dumol:
Scope
=====

RAM usage issues on Solaris 10 SPARC. Let's see if 32binaries on Solaris would alleviate this. On the other commercial UNIX'es (AIX and HP-UX) we only have 32bit binaries.

Changes
=======

Force 32bit builds on Solaris.
Replaced `if` with `case` for normalizing arch names in `paver.sh`.
Completed deps list for 32bit builds on Solaris.
Worked around an issue with forcing the new Python version for compat tests.
Also use current `paver.sh` to overwrite the one in `compat` (needed in this case because of the arch switch for Solaris).

**Things to discuss**:
  * should we force 32bit builds only for sparc (and not for amd64?)
  * should we fix `paver.sh` everywhere to accept single versions without a _default@_ prefix or should we mandate the prefix and a suffix of `:`?
 * is there any downside to overwriting `paver.sh` for the compat tests?

reviewers: @adiroiban 

Please review the changes.
Run the tests.
Check that 32bit binaries are built for all Solaris versions.

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -180,8 +180,6 @@ case $OS in
         if [ "$OS" = "solaris10u3" ]; then
             # pip doesn't like the included zlib from this old Solaris release.
             export BUILD_ZLIB="yes"
-            # GMP needs to be told that we aim for a 32bit build.
-            export ABI="32"
         fi
         # Here's where the system-included GCC is to be found.
         if [ "${CC}" = "gcc" ]; then
@@ -190,6 +188,8 @@ case $OS in
         # The location for GNU libs in Solaris, including OpenSSL in Solaris 10.
         if [ "${ARCH%64}" = "$ARCH" ]; then
             export LDFLAGS="$LDFLAGS -L/usr/sfw/lib -R/usr/sfw/lib"
+            # GMP needs to be told that we aim for a 32bit build.
+            export ABI="32"
         else
             export LDFLAGS="$LDFLAGS -m64 -L/usr/sfw/lib/64 -R/usr/sfw/lib/64"
             export CFLAGS="$CFLAGS -m64"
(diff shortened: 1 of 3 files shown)


dumol:
**needs-review**

dumol:
Struck by luck! :-]

`test_review` for above revision is all green: https://chevah.com/buildbot/builders/python-package-gk-review/builds/0

adiroiban:
32 should be only for SPARC. We not yet received complains for x86... and I don't think that we have any customers on X86/AMD64.

We are keeping the AMD64 slaves as they are fast and help with testing the Solaris differences from other Unixes.

-------------

having the config with `default` is a feature.
we should not fix paver.sh to accept a single version without `default` or os name.

There should be no issue with overwriting the paver.conf ...and having a paver.conf file is a feature as it allows us to do these type of things.

dumol:
Thanks for the review! Please notice that, with the last commit, I'm overwriting `paver.sh` in compat, not `paver.conf`, not sure if you noticed… Would that be fine too?

adiroiban:
yes. Is ok to also update paver.sh... I did not noticed at the first review.

adiroiban:
I think that paver.sh should still detect the CPU as it is.. 64bit.

Also, we should keep the python binary name as SPARC64 even if we build 32bit

The name is about the targeted CPU, and we target the modern sparc 64bit cpu.

we should build 32 only on SPARC

thanks!

needs-changes

dumol:
`paver.sh` detects the CPU correctly as `sparcv9`. But we have that sub-routine for normalizing arch names and forcing 32bit builds on some OS'es, which is a bit of kludge, but a necessary one. I tried to simplify it in this PR.

Anyway, I'm fine with forcing 32bit build on SPARC only, but I'm strongly against using `sparc64` for 32bit builds on SPARC because:

  1. We use the same convention for forcing 32bit builds on AIX. CPU is detected as `ppc64` and we force 32bit builds by using `ppc`.
  2. There are lots of place in `chevah_build` where we check for 64bit options by testing if `$ARCH` ends with `64`.

**needs-review**

PR #85, opened 2017-08-02
[#4266] Fix the build system.

adiroiban:
Scope
=====

This fix the build system for the python-package repo... and not the python distribution.

Why we got this
============

It looks like we don't have any tests before merge to test that build system is still valid

Changes
=======

Fix the brink package name and update the latest lint

reviewers: @dumol 

Check that  changes make sense


dumol:
Thanks! Have already merged and tested this in #86, it works…

**changes-approved**

PR #86, opened 2017-08-02
[#4070] FreeBSD 11 support.

dumol:
Scope
=====

FreeBSD 10 package doesn't work in FreeBSD 11.

Changes
=======

Stop trying to use `freebsd10` moniker and binaries for all FreeBSD versions.
Updated deps for FreeBSD 11.
Appended specific version in `paver.conf`.

**Drive-by changes**:
  * updated `paver.sh` and `paver.conf` from brink
  * fixed `chevah-brink` issue by merging with fixed branch
  * fixed cache validation issue in `paver.sh`.

reviewers: @adiroiban 

Please review the changes.
Run the tests.
Check out new FreeBSD 11 build slave (only one compat test fails).

### DIFF
--- a/paver.conf
+++ b/paver.conf
@@ -1,4 +1,4 @@
-PYTHON_CONFIGURATION='default@2.7.13.25cf4cf:solaris11@2.7.13.c3d8a7d:solaris10@2.7.8.c3d8a7d'
+PYTHON_CONFIGURATION='default@2.7.13.25cf4cf:solaris11@2.7.13.c3d8a7d:solaris10@2.7.8.c3d8a7d:freebsd11@2.7.13.641bbdca'
 BASE_REQUIREMENTS='chevah-brink==0.63.3 paver==1.2.4'
 BINARY_DIST_URI='http://binary.chevah.com/production'
 PIP_INDEX='http://pypi.chevah.com'
--- a/paver.conf
+++ b/paver.conf
@@ -1,4 +1,4 @@
 PYTHON_CONFIGURATION='default@2.7.13.25cf4cf:solaris11@2.7.13.c3d8a7d:solaris10@2.7.8.c3d8a7d:freebsd11@2.7.13.641bbdca'
-BASE_REQUIREMENTS='chevah-brink==0.63.3 paver==1.2.4'
+BASE_REQUIREMENTS='chevah-brink==0.63.4 paver==1.2.4'
 BINARY_DIST_URI='http://binary.chevah.com/production'
 PIP_INDEX='http://pypi.chevah.com'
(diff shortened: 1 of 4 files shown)


dumol:
**needs-review**

dumol:
Bootstrapping Python for the compat tests fail everywhere.

**needs-changes**

dumol:
Full results for all dev builders at https://chevah.com/buildbot/builders/python-package-gk-review/builds/11. 

Failures:

- [ ] new `fbsd-11` builder fails one of the compat tests, `TestSystemUsers.test_authenticateWithUsernameAndPassword_good`. More at https://chevah.com/buildbot/builders/python-package-fbsd-11/builds/11/steps/test_1/logs/stdio 
- [ ] `ubuntu-1604-arm64` fails compat testing (or is so slow that it gets killed by buildbot). As compat tests were not run properly up to now in python-package because of the cache invalidation issue, it could be that this is not a regression. More at https://chevah.com/buildbot/builders/python-package-ubuntu-1604-arm64/builds/17/steps/test_1/logs/stdio
- [ ] `hpux` fails randomly (known, filled bug), but retesting the same rev worked: https://chevah.com/buildbot/builders/python-package-hpux/builds/19

**needs-review**

paver.conf
```diff
@@ -1,4 +1,4 @@
-BASE_REQUIREMENTS='chevah-brink==0.63.3 paver==1.2.4'
-PYTHON_CONFIGURATION='default@2.7.13.4f815b5:solaris10@2.7.8.4f815b5'
+PYTHON_CONFIGURATION='default@2.7.13.25cf4cf:solaris11@2.7.13.c3d8a7d:solaris10@2.7.8.c3d8a7d:freebsd11@2.7.13.641bbdca'
+BASE_REQUIREMENTS='chevah-brink==0.63.4 paver==1.2.4'
```
adiroiban: I think that in python-package you can publish `641bbdca` for all platforms and use a simplified configuration... so that we don't have separate versions for freebsd11 and solaris11

--------

the question is whether we want to have many published version in production and get a clean config vs publishing only required changes and have a more complicated config

dumol: For disk space reasons, I would go with specific versions for specific changes and then merging them all together when there are changes that impact all platforms that require rebuilding all packages.

However, when publishing new generic versions, I would only do it after merging with  master and getting a git rev for the squashed commit. So that `git show` for that specific revision looks as informative as possible. It's a bit of a chicken and egg problem for python-package, but shouldn't be an issue for the other repos.

adiroiban: agree :) yes, on production we should only publish from master

PR #87, opened 2017-08-25
[#4299] HP-UX processor count fix.

dumol:
Scope
=====

`machinfo | grep proc | grep core | awk '{print $1}'` only works as intended on HP-UX machines with a single physical processor.

Changes
=======

After some digging, I found a command that I was able to massage into giving us the **logical** number of CPUs. However, this seems to suit HP-UX hardware, as on Mosna I got slightly better compile times with 4 threads¹ (the number of logical CPUs) compared to 2 threads² (the number of physical CPUs), when compiling `python-package` without GMP.

1. https://chevah.com/buildbot/builders/python-package-hpux-dev/builds/0
2. https://chevah.com/buildbot/builders/python-package-hpux-dev/builds/1

**Drive-by changes**:
  * re-discovered a bug with compiling GMP on HP-UX with more than 2 threads. Filled it to be reported upstream on a hacking day or something
  * count active CPUs on AIX instead of maximum available
  * some more commenting for `get_number_of_cpus()`, to not have to remember all the related gory details.

reviewers: @adiroiban 

Please review the changes.
Run `test_review` which lately also includes `hpux-dev`: https://chevah.com/buildbot/builders/python-package-gk-review/builds/18.

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -334,6 +334,9 @@ case "$ARCH" in
 esac
 if [ "${OS%hpux*}" = "" ]; then
     export PARALLEL="$JOBS"
+    # FIXME:4308:
+    # Parallel builds of GMP with more than 2 threads fail on HP-UX.
+    export PARALLEL=2
 else
     export MAKE="$MAKE -j${JOBS}"
 fi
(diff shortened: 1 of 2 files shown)


dumol:
**needs-review**

PR #88, opened 2017-09-04
[#4322] Alpine Linux support

dumol:
Scope
=====

`paver.sh` tries to use a generic Linux package in Alpine, but that fails.

Changes
=======

Identify Alpine Linux.
Check for version 3.6 or newer.
Use `lddtree -l`, as musl's ldd has issues with some Python modules.
Updated deps to accommodate this peculiar little distro.
Mark resulting `python` binary with `paxctl` for EMUTRAMP, otherwise grsec kernels will kill it, as default behaviour violates MPROTECT.
Updated `paver.conf` accordingly.

reviewers: @adiroiban 

Please review the changes.
Run the tests.
Run a server test with latest package, eg. https://chevah.com/buildbot/builders/server-alpine/builds/25

### DIFF
--- a/paver.conf
+++ b/paver.conf
@@ -1,4 +1,4 @@
-BASE_REQUIREMENTS='chevah-brink==0.63.4 paver==1.2.4'
-PYTHON_CONFIGURATION='default@2.7.13.25cf4cf:solaris11@2.7.13.c3d8a7d:solaris10@2.7.8.c3d8a7d:freebsd11@2.7.13.641bbdca:alpine36@2.7.13.a3cb2b95'
+BASE_REQUIREMENTS='chevah-brink==0.66.0 paver==1.2.4'
+PYTHON_CONFIGURATION='default@2.7.13.25cf4cf:solaris11@2.7.13.c3d8a7d:solaris10@2.7.8.c3d8a7d:freebsd11@2.7.13.641bbdca:alpine36@2.7.13.8511f0a'
 BINARY_DIST_URI='http://binary.chevah.com/production'
 PIP_INDEX='http://pypi.chevah.com'
--- a/paver.conf
+++ b/paver.conf
@@ -1,4 +1,4 @@
 BASE_REQUIREMENTS='chevah-brink==0.63.4 paver==1.2.4'
-PYTHON_CONFIGURATION='default@2.7.13.25cf4cf:solaris11@2.7.13.c3d8a7d:solaris10@2.7.8.c3d8a7d:freebsd11@2.7.13.641bbdca'
+PYTHON_CONFIGURATION='default@2.7.13.25cf4cf:solaris11@2.7.13.c3d8a7d:solaris10@2.7.8.c3d8a7d:freebsd11@2.7.13.641bbdca:alpine36@2.7.13.a3cb2b95'
 BINARY_DIST_URI='http://binary.chevah.com/production'
 PIP_INDEX='http://pypi.chevah.com'
--- a/paver.sh
+++ b/paver.sh
@@ -590,6 +590,11 @@ detect_os() {
         elif [ -f /etc/arch-release ]; then
             # ArchLinux is a rolling distro, no version info available.
             OS="archlinux"
+        elif [ -f /etc/alpine-release ]; then
+            os_version_raw=$(cat /etc/alpine-release)
+            check_os_version "Alpine Linux" 3.6 \
+                "$os_version_raw" os_version_chevah
+            OS="alpine${os_version_chevah}"
         elif [ -f /etc/rpi-issue ]; then
             # Raspbian is a special case, a Debian unofficial derivative.
             if egrep -q ^'NAME="Raspbian GNU/Linux' /etc/os-release; then
--- a/python-modules/chevah-python-test/get_binaries_deps.sh
+++ b/python-modules/chevah-python-test/get_binaries_deps.sh
@@ -22,6 +22,11 @@ elif [ "$os" = "SunOS" ]; then
     checker="ldd -L"
 elif [ "$os" = "HP-UX" ]; then
     checker="/usr/ccs/bin/ldd"
+elif [ "$os" = "Linux" ]; then
+    # objdump is more secure and portable than ldd, but more parsing is needed,
+    # as only lines that start with NEEDED are useful for deps checking.
+    # We use this as Alpine's ldd (from musl) goes nuts on some Python modules.
+    checker="objdump -p"
 fi
 
 # This portable invocation of find will get a raw list of linked libs
--- a/python-modules/chevah-python-test/get_binaries_deps.sh
+++ b/python-modules/chevah-python-test/get_binaries_deps.sh
@@ -23,10 +23,11 @@ elif [ "$os" = "SunOS" ]; then
 elif [ "$os" = "HP-UX" ]; then
     checker="/usr/ccs/bin/ldd"
 elif [ "$os" = "Linux" ]; then
-    # objdump is more secure and portable than ldd, but more parsing is needed,
-    # as only lines that start with NEEDED are useful for deps checking.
-    # We use this as Alpine's ldd (from musl) goes nuts on some Python modules.
-    checker="objdump -p"
+    if [ -f /etc/alpine-release ]; then
+        # musl's ldd has issues with some Python modules, so we use lddtree,
+        # which is forked from pax-utils and installed by default in Alpine.
+        checker="lddtree -l"
+    fi
 fi
 
 # This portable invocation of find will get a raw list of linked libs
--- a/python-modules/chevah-python-test/test_python_binary_dist.py
+++ b/python-modules/chevah-python-test/test_python_binary_dist.py
@@ -32,8 +32,6 @@ def get_allowed_deps():
             'libssl.so',
             'libutil.so',
             'libz.so',
-            'linux-gate.so',
-            'linux-vdso.so',
             ]
         # Distro-specific deps to add. Now we may specify major versions too.
         if ('rhel' in chevah_os):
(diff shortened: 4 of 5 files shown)


dumol:
**needs-review**

adiroiban:
Thanks!

I have updated buildmaster to have an alpine linux slave. Feel free to update buildmaster config.
I only wanted to see the python-package build

The slave is started with `./paver debug &` ... I did not look at the init script

It looks like it is failing https://chevah.com/buildbot/builders/python-package-alpine/builds/0

The dependencies detection need to be updated.

There are also "Error relocating" 

needs-changes

dumol:
Sorry for forgetting about the deps, I was in a bit of a rush yesterday and remembered too late about them…

Fixing this proved to be quite challenging, as Alpine's `ldd`, that comes with `musl` (the C library used instead of `glibc`), seems to be having issues with some of the Python modules. I switched to `objdump -p` as described in the man page for of the orthodox Linux `ldd`. This has allowed me to also drop the virtual Linux deps:  `linux-gate.so` and `linux-vdso.so`.

The downside is `objdump` only checks what's listed in the headers of the binaries (as does `readelf`). If you prefer we stick with `ldd` in Linux to check deps recursively, I found an alternative for Alpine Linux that works with the problematic Python modules too: `lddtree`. eg:
```
$ lddtree ./build/python2.7-alpine36-x64/lib/python2.7/site-packages/setproctitle.so
setproctitle.so => ./build/python2.7-alpine36-x64/lib/python2.7/site-packages/setproctitle.so (interpreter => none)
    libc.musl-x86_64.so.1 => /lib/libc.musl-x86_64.so.1
```
**needs-review**

adiroiban:
we can use ldd on linux and lddtree / objdump on Alpine.

we might need to run compat tests with "-v" to see exactly which tests if crashing the whole test suite :)

I am now updating brink and compat to allow running the whole suite with `-v`

I hope that by the time you will resume working on this, compat/master will be updated and ready to be used here.

needs-changes

adiroiban:
but I tried to run just the brink tests and I got a segfault when trying to send the coverage

https://chevah.com/buildbot/builders/brink-alpine/builds/5/steps/test/logs/stdio

but we can live with this for now. I have skipped coverage reporting for alpine as this is a issue with the codecov.io code.... but might releave an issue with using the OpenSSL related code.

so for python-package we might look to see wheter we can enable the pyopenssl tests as part of the test steps

adiroiban:
On Compat it fails with segfault due to PAM.

It also has another issue, in that  os.getgroups() returns duplicate groups... but we can workaround this...

and I guess that for Alpine we can completely disable PAM and system users as I guess most of the people who want to run SFTPPlus on Alpine will not want to run it as root

dumol:
I don't know about PAM, I'm surprised to see it's available on Alpine. Perhaps we should keep our options open and find a way to run with PAM on Alpine too, as hipsters run everything as root, I suppose. :-] 

Maybe tests fail because the PAM binaries compiled with glibc on Debian fail on Alpine with musl? If you agree, please fill me a ticket and I'll try to fix this at some point…

adiroiban:
For the PAM on Alpine, I am ok to ignore it for now... and do something only once we get a customer who need it.
chevah/server has config to skip PAM at all so it should work and we have a workaround.

--------

the SSL issue is important, and if possible I would like to see it fixed in this PR :)

Thanks!

dumol:
Found a way to always use OpenSSL, let me know if it's good enough. Also, should we try updating `pyOpenSSL` and/or `setproctitle` for the other OS'es using stuff from `python-modules`?

I don't mind if you work on this while I'm gone, if needed. :-]

**needs-review**

adiroiban:
I find it hard to read this diff. In https://github.com/chevah/python-package/pull/88/commits/aa73445b8e87da4e142632e5a0d86c9fb0ef35f6 I am not sure if this is vanila upstream code, or we also did some changes.

We got the python-module for legacy reason.
On modern Linux system we should use pip.

If we need to patch the upstream code, then we patch it and publish it as a new source package in pypi.chevah.com

pip install can be used to install/build from source. If we need special flags we can pass them to pip

-----------

I don't have time to work on Alpine Linux now, but I think that this should be done via pip and a source package.

-----------

I give it a try on chevah/server and it still segfaults at the same test https://chevah.com/buildbot/builders/server-alpine/builds/5/steps/test/logs/stdio

I don't know why as both python and cryptography are using the same openssl version.

-----------

The error might be due to an old version of Python Cryptography 1.9.

I see that on Alpine linux it is 2.0.3 https://pkgs.alpinelinux.org/package/edge/main/x86/py-cryptography

------------

chevah/compat is updated on master to support alpine, so we should be able to build the alpine distribution on buildslave

---------

Thanks!

needs-changes

dumol:
OK then, if the failures are present with OpenSSL too, I would rather stick with the system libs, regardless of the problem with pip. In the BSDs we also use the SSL libs in base, not OpenSSL from ports/packages. 

Unfortunately, I don't think the package version for `py-cryptography` in Alpine Linux **edge** is relevant (although I think it can't be that old). We are using the stable 3.6 release of Alpine and the included version is up to date:
```
bs1a-lnx-alpine-x64-100:~# apk info py2-crypto
py2-crypto-2.6.1-r2 description:
A Python collection of cryptographic algorithms and protocols (for python2)

py2-crypto-2.6.1-r2 webpage:
https://www.dlitz.net/software/pycrypto/

py2-crypto-2.6.1-r2 installed size:
3760128
```
So I have reverted the last two commits… With the `a3cb2b95` rev I see all of the `python-package`, `brink` and `compat` builds are green:

https://chevah.com/buildbot/builders/python-package-alpine/builds/8
https://chevah.com/buildbot/builders/brink-alpine/builds/14
https://chevah.com/buildbot/builders/compat-alpine/builds/25

**needs-review**

adiroiban:
Compat was green for a long time https://chevah.com/buildbot/builders/compat-alpine

Server is failing as soon as it needs to touch SSL/TLS :)

https://chevah.com/buildbot/builders/server-alpine/builds/14/steps/test/logs/stdio

I have not run the tests with OpenSSL.

Please look at having Alpine Linux with OpenSSL. 

It would help to include a few SSL/TLS related tests in the python-package tests, maybe run the pyOpenSSL test suite.

Thanks!

needs-changes

dumol:
Sorry, there seems to be a misunderstanding… Currently there are two packages built for Alpine at https://binary.chevah.com/production/python/alpine36/x64/, one with LibreSSL (git rev `a3cb2b95`) and the other with OpenSSL(git rev `aa73445b`).

I think you've run tests with both and there doesn't seem to be any difference. A test you've run with OpenSSL libs: https://chevah.com/buildbot/builders/server-alpine/builds/5/steps/test/logs/stdio

So, if there is nothing to be gained from compiling with third-party OpenSSL, I am for keeping to use the default LibreSSL (which is OpenSSL-based too, but forked a while ago and with some cleanups). If we are successful with LibreSSL, we could also bundle it with Solaris 10 packages, and get rid of the Python 2.7.8 dependency. LibreSSL had markedly less security issues compared to OpenSSL since forking (removing code helped). What do you say?

**needs-review**

adiroiban:
I am ok with any library :) but is important for them to be usable , right now the python-package fails for any SSL operation so I think that on whatever route we got, this still needs changes

needs-changes

dumol:
Problem is, I don't know what to do. Server builds fail similarly with both LibreSSL and OpenSSL, old cryptography version 1.9 seems a red herring (the same version appears in FreeBSD and Ubuntu builds). What next? :-]

Maybe we are missing a crucial test in `python-package`? How can we load and use the SSL libs here and they cannot be used at all in server?

adiroiban:
I don't know what is wrong... standard Python SSL seems to work, as I can download files over HTTPS with pip

I think this will just have to wait until we get a request from someone wanting to use SFTPPlus on Alpine

dumol:
Found the culprit for the signal 11 crashes, our binary was violating MPROTECT rules and the grsec-hardened default kernel of Alpine was killing it.  Would not have expected it, as the generic Linux binaries worked for me on Hardened Gentoo for years…

Anyway, after a bit of documentation it seems it's enough to enable EMUTRAMP to avoid this for a Python binary. However, it's not enough to mark the file with `paxmark`, as customary on a local system, as `tar` loses the new file attribute. So I went for an older tool, `paxctl`, that uses the program headers to mark the binary as needed for trampolines emulation under a grsec-hardened kernel. This survives packing and unpacking our package.

Have also updated corresponding `server` branch with the latest git rev of the Alpine python-package, please pull my change. A build with it: https://chevah.com/buildbot/builders/server-alpine/builds/25

Happy to finally have a working package for a hardened distro that works with the default hardened kernel and libs! :-]

**needs-review**

adiroiban:
great success :)

but, if we are adding exceptions, is our code still hardened?

Can we fix the code so that it does not need this massage?

Does the alpine own python has the same tools applied to it?


Thanks!

changes-approved 

As a followup maybe we can have public.sftpplus.com running in a docker container based on alpine? :)
What do you think?

dumol:
Yes, the code is still hardened, but the python binary is marked so that the grsec kernel knows it has to emulate trampolines for it. More at https://wiki.gentoo.org/wiki/Hardened/PaX_Quickstart#Enforce_non-executable_pages (the EMUTRAMP paragraph).

Maybe Python's interpreter can be fixed in this regard, maybe not, it's far above my head. But I observed nobody tried, so I guess it's not trivial at all, if possible.

And yes, all python binaries are marked as such in a system running a grsec kernel. Which makes me think I have disabled a corresponding protection in my Hardened Gentoo kernel (maybe because I was running VirtualBox), otherwise the generic linux python-package would not have run on my desktop.

I'm not a fan of dockerizing for security gains. More so, in this case the kernel would be the one from Ubuntu. But I am thinking on using hardened OS'es exclusively for servers running public services (eg. Yeovil, Weesp and Deag at the moment).

dumol:
@adiroiban somehow this is still in _Review required_ mode. Please approve, thanks!

adiroiban:
yes... I will look at updating the GitHub hooks so that when a comment is left, it will also sync it with the GitHub review actions

PR #89, opened 2017-10-12
[#4318] Detect Ubuntu without LBS tools.

dumol:
Scope
=====

Ubuntu detection in `paver.sh` fails if `lsb-release` package missing. Detected on Pine64 after removing `python3*` packages.

Changes
=======

Refactored Ubuntu detection to use `/etc/os-release`, present in Ubuntu 14.04 and newer, the only releases currently supported.

When older supported Linux distros (without `/etc/os-release`) are phased out (RHEL 6 and SLES 10&11), we could move to using `/etc/os-release` exclusively for distro detection in Linux (except for Arch, which is a rolling release distro and thus has no version info in that file).. 

**Drive-by changes**:
 * use `/etc/os-release` for Raspbian and Alpine Linux too
 * never build the `_locale` Python module to avoid depending on `libintl`, the GNU gettext libs
 * updated AIX deps with full paths after installing latest OpenSSL libs from IBM
 * check that `$os_version_raw` only contains digits and periods.

reviewers: @adiroiban 

Please review the changes.
Run the automated tests, eg. https://chevah.com/buildbot/builders/python-package-gk-review/builds/30
Remove `lsb*` packages and check `./paver.sh detect_os` on an Ubuntu system.
Force an unexpected character in `$os_version_raw` to test new check.

### DIFF
--- a/paver.sh
+++ b/paver.sh
@@ -593,7 +593,7 @@ detect_os() {
         elif [ -f /etc/os-release ]; then
             linux_distro=$(grep ^ID= /etc/os-release | cut -d'=' -f2)
             os_version_raw=$(\
-                grep ^'VERSION_ID=' /etc/os-release | cut -d'"' -f2)
+                grep ^'VERSION_ID=' /etc/os-release | cut -d'=' -f2)
             case "$linux_distro" in
                 "ubuntu")
                     check_os_version "Ubuntu Long-term Support" 14.04 \
--- a/paver.sh
+++ b/paver.sh
@@ -595,30 +595,28 @@ detect_os() {
             check_os_version "Alpine Linux" 3.6 \
                 "$os_version_raw" os_version_chevah
             OS="alpine${os_version_chevah}"
-        elif [ -f /etc/rpi-issue ]; then
-            # Raspbian is a special case, a Debian unofficial derivative.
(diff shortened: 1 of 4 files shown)


dumol:
**needs-review**

adiroiban:
chevah-robot did not had access to this repo, so it failed to sync with trac.

changes-approved

paver.sh
```diff
@@ -478,6 +478,7 @@ check_os_version() {
     # supported for the current OS and the current detected version.
     # The fourth parameter is used to return through eval the relevant numbers
     # for naming the Python package for the current OS, as detailed above.
+    # Beware, this function is full of bash'isms.
```
adiroiban: what is the point of this comment? the script starts with declaring that it needs bash

dumol: Not everyone uses `bash` all the time… I thought of it as a reminder to self that the following lines make repeated use of incompatible Bash extensions to the POSIX standard. So that I don't fall for checking local manuals for some stuff. :-]

If it's too much, I'll remove it.

adiroiban: we can leave it, but for me is clear that this script only works with bash... as otherwise we would just change the first line of it :)

please merge. thanks!

dumol:
With latest changes: https://chevah.com/buildbot/builders/python-package-gk-review/builds/32

PR #90, opened 2017-10-19
[#4380] Python 2.7.14, SQLite 3.20.1 and pysqlite 2.8.3

dumol:
Scope
=====

Get latest security updates and bugfixes for upstream sources.

Changes
=======

Updated Python to 2.7.14 to get latest security updates and bugfixes.
Discarded patches not required any more.
Forwarded-ported the rest of the patches.

Updated SQLite to 3.20.1 for same reason.
Updated pysqlite to 2.8.3 to have the best interface to latest SQLite. 

**Drive-by fixes**:
 * make sure `sed` invocation errors will bail out the build
 * system sqlite headers not required on Solaris 10 any more.

reviewers: @adiroiban 

Please review the commits with no imports.
Run the automated tests, eg: https://chevah.com/buildbot/builders/python-package-group-all/builds/10


dumol:
**needs-review**

adiroiban:
> Please review the commits with no imports.

which are those?

If we are using patches, why do we still need to treat the upstream as source?

Maybw we should use git attributes and set all the upstream files as binary / -diff

https://www.kernel.org/pub/software/scm/git/docs/gitattributes.html

dumol:
>> Please review the commits with no imports.
>
> which are those?

Basically, those which don't mention update/upgrading sources. There are 4 of them in total. For you convenience:

* these two have some changes of substance:
    https://github.com/chevah/python-package/pull/90/commits/846e7534e9da017d2378cdea69efc4ee8b9c8781
    https://github.com/chevah/python-package/pull/90/commits/661f3acb36a36aae0774ae34c1d7bef5a37418ba

* these two are trivial:
    https://github.com/chevah/python-package/pull/90/commits/4d94a7467307fc41ecd3b2d833ef1c8e2d066ec5
    https://github.com/chevah/python-package/pull/90/commits/9aa864d5f3b2defff2b5204403ac9af56536a461

> If we are using patches, why do we still need to treat the upstream as source?
>
> Maybw we should use git attributes and set all the upstream files as binary / -diff
> 
> https://www.kernel.org/pub/software/scm/git/docs/gitattributes.html

From what I could gather quickly (the above link has 21 pages in my maximized browser window), git attributes are a kind of per-path sticky `.gitinfo`'s. How does that help us? Thanks!

adiroiban:
I have pushed a new commit with changed attributes to see how github can handle the diff :)

Now I can get a nice diff on the command line.

```
$ git diff --stat master... | grep -v Bin
 chevah_build                                       |   9 +--
 .../chevah-python-test/test_python_binary_dist.py  |   2 +-
 src/python/chevahbs                                |  16 ++---
 src/python/crypto_threadid.patch                   |   6 +-
 src/python/hpux_fpic.patch                         |   2 +-
 src/python/hpux_nativemake.patch                   |  12 ----
 src/python/hpux_site.patch                         |   2 +-
 src/python/openbsd_random.patch                    |  12 ----
 src/python/readline_libedit.patch                  |  80 +++++++++++----------
 src/python/solaris10_paths.patch                   |   6 +-
 9274 files changed, 62 insertions(+), 85 deletions(-)
```

dumol:
Cool trick, thanks! The diff stats look clean indeed, but the diff itself still shows the names of the source files (now shown as binary), which makes it too huge to be of practical use with the many changes in this branch.

I'll upload latest test binaries and test on dedicated server branch. But why it's not enough to pass compat tests any more? It's not like there's anything I can do here, the changes are from upstream and mostly for getting up to date with latest security issues.

adiroiban:
Is ok. Please merge and push to production.

Thanks!

dumol:
Git attributes got lost, it seems… For later reference, I've used this to review my changes in this branch:
```
git diff master chevah_build src/python/chevahbs src/python/*.patch \
    python-modules/chevah-python-test/test_python_binary_dist.py
```

dumol:
`group-all` build with upload to production: https://chevah.com/buildbot/builders/python-package-group-all/builds/11 (only FreeBSD 11 failed because of a single `compat` test).

adiroiban:
yes. it looks like freebsd 11 was always failing... and was never a supported os.

In the near future I will not have time to look at freebsd 11 failes.

Should we still keep a freebsd 11 slave?

dumol:
We support FreeBSD 10.x, so I guess FreeBSD 11.x would be next in line if we continue doing that. Shouldn't take much effort, only one test is currently failing. And then I'll finally have a green python-package waterfall again! :-]

Also, if that is of any significance, I plan to use FreeBSD 11.x on my desktop at home (currently still fooling around with DragonFly BSD 5.0 with a HAMMER2 filesystem on that computer).

adiroiban:
If you are going to use FreeBSD 11 on your desktop, then I will look at fixing it.

Is a single test, but is an important one... but yes, it should not be that hard to fix :)

PR #91, opened 2017-10-26
[#4425] Use a fixed length for the git rev in the version.

dumol:
Scope
=====

With latest production Python packages, some of the packages use a git rev of 7 characters and others use 8 characters. This leads to broken tests, as the used length is not consistent among similar build slaves (being dependent of `core.abbrev` conf variable according to `git-rev-parse(1)`).

Changes
=======

Force the length of 8 for consistency.
Test the length of the git rev set in the binary.

**Drive-by change**:
 * really minor comment clarification in regards to `sqlite` libs to link to in Solaris 10.

reviewers: @adiroiban 

Please review the changes.
Run the automated tests, eg: https://chevah.com/buildbot/builders/python-package-gk-review/builds/39

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -24,7 +24,7 @@ PIP_VERSION="9.0.1"
 SETUPTOOLS_VERSION="20.3.1"
 
 # Git revision to inject into Python's sys.version string through chevahbs.
-PYTHON_PACKAGE_VERSION=$(git rev-parse --short=4 HEAD)
+PYTHON_PACKAGE_VERSION=$(git rev-parse --short=8 HEAD)
 # The ID of the redistributable version used on Windows.
 REDISTRIBUTABLE_VERSION="9.00.30729.6161"
 
--- a/chevah_build
+++ b/chevah_build
@@ -24,7 +24,7 @@ PIP_VERSION="9.0.1"
 SETUPTOOLS_VERSION="20.3.1"
 
 # Git revision to inject into Python's sys.version string through chevahbs.
-PYTHON_PACKAGE_VERSION=$(git rev-parse --short HEAD)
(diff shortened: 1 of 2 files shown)


dumol:
Please let me know if you are fine with a length of 4 or you have other ideas… Thanks!

**needs-review**

adiroiban:
Thanks for working on this.

I don't know how and if the current change would help :)

I would say that in order to get consistent results we should have minimum 8 characters :) extra safety 

I don't know how we ended up with git running for the same revision and getting different values... and I don't know why min 4 will do the job.

I guess what when we do shallow clones (depth=1) for git is hard to know if there are conflicts.

We are forcing 4 characters now, but in the future , python package repo might grow and have conflicts in 4 characters and then git-rev will use 5 even if we ask for 4.

I am not convinced that this will fix it, and I would say that we should go for a value higher than 4 ... maybe 8?

needs-changes

dumol:
The current change does help. The problem is solved with any length, as the variability was killing us (sometimes the git rev was 7 chars longs, other times it was 8 chars long).

In regards to the other points you've raised, they are valid, although not really affecting us, I think. Collision wouldn't be much of a problem as the older binary packages will be overwritten (save the special case of older git clones trying to download a newer package). And if the minimum length is raised upstream, we could raise it here too, in a similar commit.

But all in all, I don't mind a length of 8, will follow up with a patch implementing just that… Thanks for the review!

dumol:
Implemented suggested changes.

Added a new test for the length of the git rev set in the binary.

**needs-review**

dumol:
Maybe we should use 7, as GitHub does? Those short 7 chars git rev are visible all over the place around here…

PR #92, opened 2017-10-30
[#4437] Minor /etc/os-release fixes.

dumol:
Scope
=====

`VERSION_ID` is sometimes missing from `/etc/os-release`.  For example on my old Gentoo installation that I keep around at home. Result:
```
$ ./paver.sh
./paver.sh: line 603: VERSION_ID: unbound variable
```

Also, Ubuntu non-LTS versions are now rejected, while we used to equate them with generic Linux before introducing support for `/etc/os-release`.

Changes
=======

Only use `VERSION_ID` from `/etc/os-release` on supported distros, for now: Ubuntu, Raspbian and Alpine. This variable is not set on some rolling-release distros according to https://www.freedesktop.org/software/systemd/man/os-release.html.

Use generic Linux binaries on non-LTS Ubuntu versions.

reviewers: @adiroiban

Please review the changes.
Run `paver.sh` on a rolling-release distro such as Gentoo.
Run `paver.sh` on a non-LTS Ubuntu version.
Run the automated tests.

### DIFF
--- a/paver.sh
+++ b/paver.sh
@@ -600,10 +600,10 @@ detect_os() {
         elif [ -f /etc/os-release ]; then
             source /etc/os-release
             linux_distro="$ID"
-            os_version_raw="$VERSION_ID"
             distro_fancy_name="$NAME"
             case "$linux_distro" in
                 "ubuntu")
+                    os_version_raw="$VERSION_ID"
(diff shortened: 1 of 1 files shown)


dumol:
**needs-review**

PR #95, opened 2017-12-06
[#4438] Use /etc/os-release for Arch Linux too.

dumol:
Scope
=====

Now that we don't use VERSION_ID for all distros where we count on `/etc/os-release`, Arch as rolling-release distro should fit among them as well.

Changes
=======

Stop using `/etc/arch-release` in favour of `/etc/os-release`.

**Drive-by change**:
  * updated `paver.conf`

reviewers: @adiroiban 

Please review the changes.
Check that Arch is detected properly by `paver.sh`.
Run the automated tests.


dumol:
**needs-review**

PR #101, opened 2018-04-02
[#4836] SQLite upgrade to 3.23.0.

dumol:
Scope
=====

Build python-package with a more recent SQLite, as the version we are currently using, 3.20.1, is affected by a few security issues: ​https://www.cvedetails.com/vulnerability-list/vendor_id-9237/Sqlite.html
​https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=sqlite

Note that on Windows we are at the mercy of the upstream Python packagers and the version they include, 3.8.11.0, is affected by even more serious issues.

Changes
=======

Updated the included SQLite sources and incremented SQLITE_VERSION accordingly.

Added an up-to-date SQLite 32-bit Windows DLL in `src/sqlite` and used it to overwrite the DLL included in Python's Windows installer.

reviewers: @adiroiban 

Please review the changes.
Run the automated tests.
Make sure the updated Windows DLL works as expected.


dumol:
**needs-review**

dumol:
Both the sources and the 32bit DLL were downloaded from upstream sqlite.org site. I thought it goes without saying, but it seems that's not the case, so I documented it  to be clear. Could be helpful in some other ways, as there are plenty of downloads to choose from on the sqlite.org download page.

However, upstream Python packagers were not using the same source. From what I can tell, they build the SQLite DLL themselves, possibly with Visual Studio, as apparently they are not using GCC as the SQLite maintainer does.

PR #102, opened 2018-04-03
[#4837] cryptography updates.

dumol:
#Scope
=====

Version 1.9 we are currently using on most OS'es is not known to have security issues, but it bundles OpenSSL 1.1.0f on Windows and OSX / MacOS, which does have a number of issues: ​https://www.cvedetails.com/vulnerability-list/vendor_id-217/product_id-383/version_id-223239/Openssl-Openssl-1.1.0f.html.

More so, newest pyOpenSSL requires newer `cryptography`. pyOpenSSL has fixed in latest versions bugs with a security impact. More at https://github.com/pyca/pyopenssl/blob/master/CHANGELOG.rst

Changes
=======

Updated `cryptography` for SLES 11 (and possibly RHEL 5) to latest version with support for OpenSSL 0.9.8, namely version **1.3.4**.

Updated `cryptography` to latest version **2.2.2** for all OS'es using 1.9.

Switched to using cffi-based modules for AIX, HP-UX and Solaris 11 with the following `cryptography` versions:
  * pristine **2.2.2** for AIX, with a minor patch for Python to find `ld_so_aix` & co.
  * slightly patched **2.2.2** for HP-UX (minor problems with CFLAGS and C syntax)
  * **2.2.2** / **1.8.2** / **1.6** for Solaris 11, depending on OpenSSL version and availability of EC crypto, more in https://github.com/chevah/python-package/pull/102/commits/6fc61a51dd32a18e22cf2f4a2e45d43295c06e70. A bit of pruning for compile flags is needed for 2.2.2.

**Drive-by fixes**:

  * updated `pip` and `setuptools` to latest stable versions: 9.0.3 and 39.0.1
  * updated `cffi` to latest version, 1.15.2, to avoid reinstalling when building `cryptography`'s and related issues on exotic platforms
  * happily hacked in support for OpenBSD 6.3 and LibreSSL 2.7.x
  * updated SQLite sources and Windows DLL to fix a nasty bug found upstream
  * updated `scandir` to latest version, as 1.5 PyPI archive was not handled well in the Unix world (CR+LF endings). `scandir` tests now pass on Debian Linux and Solaris 10 AMD64 too
  * call `ioscan` with full path on HP-UX to account for a $PATH with no `/usr/sbin`
  * updated `paver.conf` for Ubuntu 18.04
  * updated Solaris 11 installation on Jacu from 11.1 to 11.2 to match the Solaris 11 AMD64 installations. OpenSSL version was upgraded from 1.0.0j to 1.0.1h. 

reviewers: @adiroiban 

Please review changes in `chevah_build` and `src/python`, everything else is related to updating upstream sources.
Run the automated tests.
Check out cffi-based modules in AIX / HP-UX / Solaris 11, especially `cryptography`.
Check failed `compat` tests on Solaris 11 SPARC.


dumol:
A significant chunk of this has been done during Happy Hacking Day (OpenBSD-related stuff) and in my spare time (support for older AIX versions and unsuccessfully hacking on `libedit` for Solaris).

AIX 6 slave can't build this successfully, but I'll work more on the "new" installation in Foi in my spare time. On AIX 5.3 things are smooth.

Beware of the Solaris 11 SPARC update to 11.2, which breaks `compat`. More stuff in the PR description.

**needs-review**

adiroiban:
I am looking at this, but I  try to think how to approach it.

I am planning to use this in chevah/server and see if test pass... and with that we start using the latest versions.

dumol:
This is work in progress, I still have to update at least pyOpenSSL and Python (2.7.15 was launched a few days ago). Please review the changes in `chevah_build` and `src/python/chevahbs` and let me move on with the security-related updates until I reach a stage where it makes sense to start using this into the `server` repo and roll the security fixes in a new SFTPPlus release. Thanks!

chevah_build
```diff
@@ -661,7 +707,7 @@ command_test() {
             # available, but they need to be requested for the tests to pass.
             export LC_ALL=en_US.UTF-8
             execute $LOCAL_PYTHON_BINARY \
-                ../python-modules/python-scandir-1.5/test/run_tests.py
+                ../python-modules/python-scandir-1.7/test/run_tests.py
```
adiroiban: maybe we can store the scandir version in a varialble

chevah_build
```diff
@@ -305,7 +351,7 @@ if [ "${OS%solaris*}" = "" ]; then
         # Required for linking to libedit, which has a simpler configure setup.
         export CFLAGS="$CFLAGS -xcode=pic32"
     fi
-elif [ "${OS%freebsd*}" = "" ]; then
+elif [ "${OS%freebsd*}" = "" -o "${OS%openbsd*}" = "" ]; then
```
adiroiban: maybe is best to have a separate elif for openbsd as the next comment only talks about freebsd 10

chevah_build
```diff
@@ -258,6 +292,18 @@ case $OS in
...
         export MACOSX_DEPLOYMENT_TARGET=10.12
         ;;
+
+    openbsd*)
+        export CC=clang
+        export CXX=clang++
+        # Fix from upstream, newer cryptography versions should work as-is.
+        PIP_LIBRARIES="\
+            setproctitle==1.1.10 \
+            cryptography==2.2.2chevah2 \
```
adiroiban: I don't understand these PIP_LIBRARIES set.
For example here we have 2.2.2chevah2 and on other places we have 2.2.2chevah3

I would to have something like a LATEST_LIBRARIES variable defined and documented at the top of the script, and then here only have `PIP_LIBRARIES=LATEST_LIBRARIES`

chevah_build
```diff
@@ -268,12 +314,12 @@ case $OS in
         export BUILD_LIBEDIT="no"
         ;;
     rhel5|sles11)
-        # Up-to-date cryptography not working on these...
+        # cryptography 1.3.4 is latest version with support for OpenSSL 0.9.8.
```
adiroiban: I would also likes to see something like `PIP_LIBRARIES=OPENSSL_098_LIBRARIES`

Where OPENSSL_098_LIBRARIES is documented at the top, and consider whether we can use it on OSX.

chevah_build
```diff
@@ -268,12 +314,12 @@ case $OS in
...
         PIP_LIBRARIES="\
             setproctitle==1.1.10 \
-            cryptography==1.3.1 \
+            cryptography==1.3.4 \
             pyOpenSSL==16.0.0 \
```
adiroiban: Can we use `pyOpenSSL==17.0.0` here? Maybe we should add a comment why it can't be used

chevah_build
```diff
@@ -208,11 +204,44 @@ case $OS in
...
+            elif [ -f /usr/include/openssl/cmac.h ]; then
+                # 1.8.2 works with the crippled OpenSSL 1.0.1 in Solaris 11.2.
+                PIP_LIBRARIES="\
+                    setproctitle==1.1.10 \
+                    cryptography==1.8.2 \
+                    pyOpenSSL==17.0.0 \
+                    scandir==1.7 \
+                    "
+            else
+                # 1.6 still supported OpenSSL 1.0.0, the default in 11.0/11.1.
```
adiroiban: Do we need to support all theses solaris 11 versions?

chevah_build
dumol: OK.

chevah_build
dumol: Actually this is about Clang, which is the compiler used in both FreeBSD and OpenBSD nowadays. Will update the comment to reflect this. Thanks for spotting it!

I think after getting rid of support for OS X 10.8, it would be worth checking if changing the condition to look for `$CC = "clang"` works for MacOS too.

chevah_build
dumol: There are multiple fixes in the 2.2.2chevah versions. I tried to document them in the comments for every occurrence. Eg., for 2.2.2chevah2, you can read in the comment above that it includes a fix from upstream that will be included in 2.2.3. 

Obviously, this is a fix specific for this OS: OpenBSD. Cryptography 2.2.2 (and I think any older version too) will not work on OpenBSD 6.3 because of an incompatibility with the LibreSSL libs bundled with OpenBSD 6.3. 

So it's not possible to use `LATEST_LIBRARIES` **unless** you are willing to use these fixes for all OS'es. I have chosen to use pristine upstream sources wherever possible (so everywhere except OpenBSD, HP-UX and Solaris 11).

chevah_build
dumol: I have another ticket for that: https://trac.chevah.com/ticket/4838. It's actually the reason for updating cryptography in the first place, as it's a security bug.

chevah_build
dumol: OK, no problem. Although this is the only place where these libs are used.

chevah_build
dumol: My efforts were into having things work consistently, that's why I upgraded Jacu from 11.1 to 11.2. Obviously, we can't support all versions of Solaris 11, like we don't support building with GCC on Solaris. But I chose to document these conditions, because you never know when/if you are going to need them again… Or maybe someone else will need them, as this is a public repo.

chevah_build
dumol: But actually yes, in this particular case, I think pyOpenSSL version 16.0.0 is the latest version with support for cryptography 1.3.4. More work on documenting these will come in the aforementioned ticket.

chevah_build
dumol: Have actually generalized the idea, as it makes it easier to see where we use custom libs. So:
```
# Python modules versions to be used everywhere possible.
CFFI_VERSION="1.11.5"
PYCRYPTO_VERSION="2.6.1"
PYOPENSSL_VERSION="17.0.0"
CRYPTOGRAPHY_VERSION="2.2.2"
SETPROCTITLE_VERSION="1.1.10"
SCANDIR_VERSION="1.7"
```
And then, where's a change you see something along the lines of:
```
# cryptography 1.3.4 is latest version with support for OpenSSL 0.9.8.
PIP_LIBRARIES_OPENSSL_098="\
    setproctitle==${SETPROCTITLE_VERSION} \
    cryptography==1.3.4 \
    pyOpenSSL==16.0.0 \
    scandir==${SCANDIR_VERSION} \
    "
```

chevah_build
adiroiban: is hard to follow the commit in here.
I would like to see the changes made in a brank in our fork of cryptography and have those versions in pypi.chevah.com 
https://github.com/chevah/cryptography/branches

PR #104, opened 2018-05-08
[#4908] Python 2.7.15 update.

dumol:
#Scope
=====

Python 2.7.15 was released a few days ago and there are some security fixes in there (including an updated copy of `expat`), according to ​https://github.com/python/cpython/blob/2.7/Misc/NEWS.d/2.7.15rc1.rst, so I'm treating this as a security issue.

Changes
=======

Updated Python sources and Windows installer to 2.7.15 from upstream.

Updated our patches accordingly.

**Drive-by changes**:

 * `nis` modules is disabled, as we don't use it and it links to lots of libs (GSS-API, Kerberos etc.)
 * a second AIX-specific hack for `ld_so_aix` woes, as upstream fixes from Python's bug tracker do not cover all scenarios
 * cherry-picked Ubuntu 18.04 updates from #103
 * reverted `chevah-keycert` tests
 * excepted unsupported platforms such as `solaris11-sparc` and `freebsd11` from running `compat` tests.
 * minor tweaks for Solaris 10.

How to try and test the changes
===============================

reviewers: @adiroiban 

Review changes to master in `chevah_build`, `functions.sh` and `src/python/*.patch`.
Think of a better solution for AIX's `ld_so_aix` woes?
Run the automated tests.

PR #105, opened 2018-06-01
[#4952] Build gmpy2 (or gmpy) module.

dumol:
#Scope
=====

We need gmpy (version1 ​https://pypi.org/project/gmpy/#files) to speed ssh operations
```
$ ./build-ubuntu1604-x64/bin/python 
Python 2.7.14 (620df8b0, Nov  9 2017, 16:16:50) 
[GCC 5.4.0 20160609] on linux2
>>> import gmpy
  File "<stdin>", line 1, in <module>
ImportError: No module named gmpy
>>> 
```
the tests in python package needs to be updated to have import gmpy

we can build it in python-package or create wheels for each platform

Later we should migrate to gmpy2... I have start looking at converting the code from gmpy2 to gmpy, but meanwhile we should have gmpy functional

Changes
=======

`gmpy` is unmaintained, has been superseded by `gmpy2`.

Build `gmpy2` in a dedicated step, as with `pysqlite`, because we need to build it without MPFR and MPC (we only need MP with GMP at the moment). 

But use the official wheel for Windows (which has MPIR included instead of GMP).

I've added a basic test for `gmpy2` for what we use, as discussed on IRC.

Have tried adding more tests from the `test*` sub-dirs, but most of them have multiple broken tests on all platforms, see for example the patch and first test phase for https://chevah.com/buildbot/builders/python-package-win-xp-dev/builds/88. The only fully-functional test seems to be `test3/test_mpz_args.py`, not sure if we want this, it's not quite related, AFAICT.

For HP-UX was only able to get a functional `gmpy2` when building 64bit binaries, but then some other stuff broke: a minor module in `cryptography`, `setproctitle` and `_scandir` (but not `scandir`). Have tried for some hours, mostly in my free time, to fix either the 64bit issues or `gmpy2` for 32bit builds, but failed.

So on HP-UX we are using `gmpy` instead of `gmpy2`.

**Drive-by fixes**:
  * use `aCC -Ae` insted of `cc` on HP-UX to make it possible to use more than 2 threads when compiling GMP
  * link with `aCC -b` on  HP-UX to allow for CFLAGS not supported by default `ld`.

reviewers: @adiroiban 

Please review changes in `chevah.sh` and `python-modules/chevah-python-test/test_python_binary_dist.py`.
Run the automated tests.


dumol:
**needs-review**

dumol:
@adiroiban, `2.7.15.ba61f65` binaries uploaded to testing for all platforms and arches. Hope everything will turned out fine in `server`.

adiroiban:
Just a note

@dumol for hpux. I tried to take builds from older PRs and test HPUX part by part.
For example I looked at https://github.com/chevah/python-package/pull/104 and tried a rebuild but it failed now

* Failed rebuild now https://chevah.com/buildbot/builders/python-package-hpux-dev/builds/255
* Success build in 9 may https://chevah.com/buildbot/builders/python-package-hpux-dev/builds/102

I was expecting it to fail maybe at pyopenssl, but it looks like that is not the case.

In an ideal world, we could look at any past build which was green and be rebuilding it we should also get a new green build. We can discuss over IRC what option we have.

Thanks!

dumol:
Thanks for reviewing this!

HP-UX rebuilds fail roughly 2 out 3 times because of the issues in https://trac.chevah.com/ticket/4205. You should just retry until successful, if the failure is one of those listed there.

However, I'm confident we don't have to backtrack that much for HP-UX. I'll try to get a better build here. I think the build improvements in https://github.com/chevah/python-package/pull/105/commits/4d12876e966e5a331a069bea0adcd016965a40e2 are causing the segmentation faults in server, most probably the flag for position-independent code.

As for AIX 7, I've tried several months ago to rebuild on AIX 7.1 with OpenSSL 1.0.1e (there are several releases from IBM with this version, nothing worked for me). The issues are documented in Trac, ticket  https://trac.chevah.com/ticket/4797, and the associated PR.

For AIX, **if** there's still a need for supporting OpenSSL 1.0.1 (for which IBM dropped support an year and a half ago), my suggestion would be to do things as in RHEL 7, but with an older Python package, not recommended for anything else but as a way to upgrade to latest stuff. No one should be using that unsupported OpenSSL version in AIX any more.

adiroiban:
Thanks for AIX we need to communicate with the customers and see what they want.

For now,  I will put AIX 7.1 with old OpenSSL in unsupported on chevah/server

PR #106, opened 2018-06-06
[#4833] Build libffi only when needed.

dumol:
Scope
=====

Our bundled `libffi` is used twice:

  1. by Python if using `--with-system-ffi` (on AIX and HP-UX)
  2. by CFFI (on all platforms where we use pip/cffi/cryptography).

Stop building `libffi` unnecessarily on OS'es which bundle and support it. Linking to OS-provided `libffi` means one less liability from a security point of view.

Changes
=======

Build CFFI against OS libs where this is possible (everywhere but AIX and HP-UX).

But also on the BSDs, where `libffi` is not included in the base system and has to be installed as a package. I'm ambivalent about this one though, as you can see from the history of changes.

Systems not building CFFI are not affected (currently Windows and Solaris 10).

Systems not tested, but for which I deduced required deps: Solaris 11 SPARC 64bit (as we force 32bit builds) and Raspbian 7.

**Drive-by fixes**:
  * lower `MACOSX_DEPLOYMENT_TARGET` to 10.11 in preparation for dropping support for OS X 10.8 - 10.10
  * cosmetic changes for how we document and check for needed packages on some Linux distros.

reviewers: @adiroiban 

Please review changes.
Run the automated tests.

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -350,6 +350,7 @@ case $OS in
     openbsd*)
         export CC="clang"
         export CXX="clang++"
+        export BUILD_LIBFFI="yes"
         # Fix from upstream, newer cryptography versions should work as-is.
         PIP_LIBRARIES="\
             setproctitle==${SETPROCTITLE_VERSION} \
@@ -361,8 +362,7 @@ case $OS in
     freebsd*)
         export CC="clang"
         export CXX="clang++"
-        # CFFI build doesn't find libffi.
-        export LDFLAGS="$LDFLAGS -L/usr/local/lib"
+        export BUILD_LIBFFI="yes"
         ;;
     linux*)
         # We don't compile libedit for generic Linux builds because it links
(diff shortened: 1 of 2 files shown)


dumol:
**needs-review**

dumol:
Thanks for the review, I will merge this in master and then into the gmpy PR, #105. 

One thing I forgot is to lower the deployment target for MacOS to 10.11 to cover all versions used by Ixmal, once we drop OS X support. Will do this as a drive-by here.

> I am ok to keep libffi as an external dependency in FreeBSD.

I thought the same in the beginning (see https://github.com/chevah/python-package/pull/106/commits/f725362e9707b3b55f4d4eeead52686818e2f82f), but I've changed my mind since then. And the more I think about it, the more I lean to the side of not using packages on the BSDs. I'm very sure it's the right thing to do on OpenBSD, not that sure on FreeBSD, but better to keep things on the safe side.

> Please double check that once this is used in chevah/server we update the documentation to note the libffi is require for freebsd
> […]
> In this list I don't see the change for Arch / Alpine.

Because we don't do package verifications for those OS'es, it's only something I planned for future happy hacking sessions. But you can see I have updated the list for Alpine on Wednesday in my ticket: https://trac.chevah.com/ticket/4773. When done with this, I can do the same for Arch Linux. If we still support it by then, that is. :-]

As this doesn't appear in my `needs_merge` list on Trac:

**changes-approved**

adiroiban:
> but better to keep things on the safe side.

ok. you are our only customer/stakeholder for openbsd/freebsd so I will go with your suggestions :)

PR #107, opened 2018-06-15
[#4205] HP-UX build fixes.

dumol:
Scope
=====

HP-UX python-package builds sometimes fail on Mosna.

Initially, this seems to be related to Mosna specifically, but further testing focused on this specific issue revealed the root cause for this to be parallel building, especially for GMP, but sometimes for Python as well.

Changes
=======

Use `gmake` for GMP and Python on HP-UX. This means no parallel builds for these two components on HP-UX.

Switch back to using the regular C compiler on HP-UX, instead of the C++ compiler in _C mode_. This was only necessary for building GMP with more than 2 threads.

**Drive-by change**:
  * updated `paver.conf` with latest stuff.

reviewers: @adiroiban 

Please review changes.
Run the tests. You'll have to do 10 in a row to make sure these failures are gone.

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -310,16 +310,14 @@ case $OS in
         ;;
     hpux*)
         # GCC and the bundled /usr/ccs/bin/cc will fail with pycrypto.
-        # We prefer "aCC -Ae" over "cc" because it does parallel builds better.
-        export CC="/opt/aCC/bin/aCC -Ae"
-        export CXX="/opt/aCC/bin/aCC"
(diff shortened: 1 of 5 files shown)


dumol:
**needs-review**

dumol:
Thanks for the review!

An interesting related observation is that we use 32 builds for all Unices, except Solaris 10 x64. But only in HP-UX we use a `64` suffix for 32bit builds (meaning `ia64` here for Itanium).

adiroiban:
Well... we can't use ia32 as that is x86 :)

PR #109, opened 2018-07-04
[#4773] Package verification for all Linux distros, check essential commands everywhere else.

dumol:
Scope
=====

Following the work in Trac ticket #4712, we don't install/uninstall packages anymore when building Python, but we check for required packages on some of the systems.

We could add Alpine to that list of systems.

Changes
=======

Check required packages in Alpine with `apk` (and in Arch Linux with `pacman` while at it).

Check that essential commands are present on OS'es where we don't check for installed packages: Windows, the Unices, OSX/macOS, the BSDs and generic Linux.

**Drive-by changes**:
  * also check for `git` and `patch` everywhere as they are actually required
  * fixed a bug that prevented package verification on SLES 11 in last revisions
  * various related minor fixes: comments, outputs, code syntax.

reviewers: @adiroiban 

Please review changes.
Run the automated tests for all OS'es, eg. https://chevah.com/buildbot/builders/python-package-group-all/builds/16.
Add a common fake required package and/or command and check that the build is not started.

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -114,7 +114,7 @@ OS=`cut -d' ' -f 3 DEFAULT_VALUES`
 ARCH=`cut -d' ' -f 4 DEFAULT_VALUES`
 
 # List of OS packages required for building Python/pyOpenSSL/cryptography etc.
-COMMON_PKGS="gcc make m4 automake libtool texinfo patch"
+COMMON_PKGS="gcc make m4 automake libtool texinfo git patch"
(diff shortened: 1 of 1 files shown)


dumol:
**needs-review**

dumol:
@adiroiban, maybe you lost track of this? I would like to merge this in master if changes are fine by you. Thanks!

dumol:
Thanks!

`command -v` is something from POSIX, mandatory since the 2008 revision, it seems. But that is not relevant here, as we use Bash anyway.

PR #108, opened 2018-07-03
[#4970] PyCrypto update from master.

dumol:
Scope
=====

Ditto, to fix ​https://nvd.nist.gov/vuln/detail/CVE-2013-7459, as per ​https://github.com/dlitz/pycrypto/issues/176 and other issues fixed since latest stable release.

Prior discussions on this at https://trac.chevah.com/ticket/4788.

Changes
=======

Uses latest `pycrypto` sources from master.

Fixed a minor build issue on Solaris.

Fixed build issue on macOS / OpenBSD / FreeBSD 11 with Clang 4.x using a backport fix for `libtom` 17.0 from https://github.com/Homebrew/homebrew-core/pull/3922.

**Drive-by fixes**:
  * only set specific CFLAG for `libedit` on Solaris when building `libedit`
  * fix deps test on Ubuntu 18.04 to account for latest `ldd`'s output on statically linked binaries
  * don't bother building old `cryptography` 1.8.2 on OS X, we have a working 2.2.2 wheel from upstream.

How to try and test the changes
===============================

reviewers: **WIP**

PR #110, opened 2018-07-12
[#4985] PyCryptodome instead of PyCrypto.

dumol:
Scope
=====

To fix ​​https://nvd.nist.gov/vuln/detail/CVE-2013-7459 and other known and unknown vulnerabilities from `pycrypto`, which has been unmaintained for several years now. Prior discussions on this in Trac ticket #4788.

There is also #108  for updating `pycrypto` from master, but knowing `pycryptodome` is (almost) a drop-in replacement, this might be easier. And it's definitely the better option.

Changes
=======

As this is not quite a drop-in replacement, it's not feasible to use `pycrypto` and `pycryptodome` side by side, so all traces of `pycrypto` have been removed and `pycryptodome` was fixed to build on all supported platforms, with help from upstream. The fixes have been incorporated in a new stable version : 3.6.4.

An extensive test suite from upstream is run an all platforms, except HP-UX, for now. This requires quite some RAM, so slaves building this have been given at least 1GB of RAM, but typically 1.5GB. It also increases build times considerably, especially on Unices, because their proprietary compilers have no support for 128bit integers. Build times for SPARC have doubled!

There are currently minor issues with PyCryptodome's builds on RHEL 6 and SLES 11 (with or without the Security Module). Working with upstream on fixing these issues one by one. For now we use the upstream wheel on these platforms for the current version. The way we do it should make it painfully obvious when upgrading the generic version (defined at the beginning of `chevah_build`), to avoid these platforms being left behind when incrementing the generic version.

There are major issues on HP-UX with running the PyCryptodome tests (segmentation faults), but they seem caused by `libffi`/`cffi`. Could be related with the HP-UX crashes we get lately. To be investigated and hopefully fixed in another ticket.

**Drive-by changes**:
  * Only set specific CFLAGS for `libedit` on Solaris when building `libedit`.
  * Fix deps test on Ubuntu 18.04 to account for latest `ldd`'s output on statically linked binaries.
  * Don't bother building old `cryptography` 1.8.2 on OS X, we have a working 2.2.2 wheel from upstream. In OS X, OpenSSL 0.9.8 is only used for the standard `ssl` module now. However, stop using that wheel on MacOS, where we want OpenSSL updates through Homebrew.
  * Use the upstream Linux generic `cryptography` 2.2.2 wheel on SLES 11 as well, but only on SLES 11, as we don't want cryptography's baked-in OpenSSL to override the locally installed OpenSSL on other Linux distros, not even on SLES 11 SM.
  * Always use `pip` to get PyCryptodome, for more flexibility in using wheels on some Linux platforms (see above comments on RHEL 6, SLES 11 and SLES 11 SM).

reviewers: @adiroiban 

Please review changes in `chevah_build`, `python-modules/chevah-python-test/test_python_binary_dist.py`  and `src/libffi/chevahbs`.
Run the automated tests.
Start migrating to `pycryptodome`? This seems to be the definitive guide: http://pycryptodome.readthedocs.io/en/latest/src/vs_pycrypto.html


dumol:
**needs-review**

chevah_build
```diff
@@ -342,10 +347,12 @@ case $OS in
...
         export CFLAGS="$CFLAGS -mmacosx-version-min=10.8"
         export MACOSX_DEPLOYMENT_TARGET=10.8
-        # OS X 10.8 not supported by cryptography 1.9, so we use 1.8.2.
+        # OS X 10.8 not supported by cryptography 1.9, so you'll normally use
+        # 1.8.2, but cryptography 2.2.2 has a working wheel for MacOS 10.6+.
         PIP_LIBRARIES="\
             setproctitle==${SETPROCTITLE_VERSION} \
-            cryptography==1.8.2 \
+            pycryptodome==${PYCRYPTODOME_VERSION} \
+            cryptography==${CRYPTOGRAPHY_VERSION}upstream \
```
adiroiban: what is this upstream?

dumol: As described in the PR description: __[…]For now we use the upstream wheel on these platforms for the current version. The way we do it should make it painfully obvious when upgrading the generic version (defined at the beginning of chevah_build), to avoid these platforms being left behind when incrementing the generic version._

And: _Don't bother building old cryptography 1.8.2 on OS X, we have a working 2.2.2 wheel from upstream. In OS X, OpenSSL 0.9.8 is only used for the standard ssl module now. However, stop using that wheel on MacOS, where we want OpenSSL updates through Homebrew._

src/libffi/chevahbs
```diff
@@ -54,14 +54,17 @@ chevahbs_install() {
...
 
-    # libffi installs its headers in $PREFIX/lib/libffi-$VERSION/include,
-    # instead of $PREFIX/include, so we copy them to the default location.
-    execute mkdir -p $INSTALL_FOLDER/include
-    execute cp $INSTALL_FOLDER/lib*/libffi-*/include/* $INSTALL_FOLDER/include/
-    # On some systems libffi is installed in lib64/ and then cffi is
-    # searching for it in lib/ (this affects RHEL 5-7 and SLES 11-12).
-    if [ -d $INSTALL_FOLDER/lib64 ]; then
-        execute cp $INSTALL_FOLDER/lib64/* $INSTALL_FOLDER/lib/
+    if [ -z "$(ls -A $INSTALL_FOLDER/include)" ]; then
```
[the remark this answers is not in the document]
adiroiban: I see that this is due to hpux... but since hpux is broken, I think that we should leave this for a separate PR and not have this change here

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -646,9 +647,10 @@ def main():
 
     try:
         import Crypto
-        print 'PyCrypto %s' % (Crypto.__version__,)
+        pycrypto_version = Crypto.__version__
```
adiroiban: Instead of the compatible "Crypto", I think that we should install `pycryptodomex` and have import `import Cryptodome`

In this way, we are sure that we have PyCryptodome.
Otherwise this test will pass even when pycrypto is installed

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -744,13 +746,6 @@ def main():
             sys.stderr.write('"setproctitle" missing.\n')
             exit_code = 7
 
-        try:
-            from Crypto.PublicKey import _fastmath
-            exit_code = egg_check(_fastmath) | exit_code
-        except:
-            sys.stderr.write('Crypto.PublicKey._fastmath missing. No GMP?\n')
```
adiroiban: there is no  fastmath support in cryptodome?

adiroiban:
I have pushed the .gitattributes, but GitHub is ignoring the `.gitattributes` file.
So other than having a nicer list with `$ git diff --stat master... | grep -v Bin` there is not much use for .gitattributes

As noted here https://stackoverflow.com/questions/7607739/gitattributes-not-making-a-difference-trying-to-skip-files-when-using-git-diff `git difftool` is also ignoring .gitattributes

src/libffi/chevahbs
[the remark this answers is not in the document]
dumol: HP-UX is broken in master and I started to work on it here, hoping for a quick fix, hence the above changes that use a RC release of `libffi` for HP-UX. This work is continued in branch `5017-hpux-cffi-crash`.

python-modules/chevah-python-test/test_python_binary_dist.py
dumol: This is not a test, it should just output the version. The tests for PyCryptodome are in `chevah_build. PyCrypto is removed completely in this branch anyway.

python-modules/chevah-python-test/test_python_binary_dist.py
dumol: The tests for PyCryptodome are in `chevah_build`. There is support for using GMP for faster math.

dumol:
> I prefer to have the hpux fix in a separate PR.

I'm working on HP-UX in another branch, derived from this: `5017-hpux-cffi-crash`. Using the latest `libffi` RC on HP-UX I see as a non-issue, because HP-UX is broken in master.

> Also, I think that we can leave rhel5/6 and sles 11 for a separate PR.

Do you care to elaborate? PyCryptodome's maintainer has just started his summer vacation and won't come back for 6 weeks. Even then, there's no guarantee he will solve these problems with priority. The issue on RHEL 6, for example, does not replicate with the distro's Python.

> As a start I want to have an ubuntu 16.04 with pycryptodome so that I can work on parting chevah-keycert

I don't like goal posts moving all the time. First, PyCryptodome is not good, it's not fully compatible. Then maybe it's good, give it a try. No, it's no good for us particularly, work on PyCrypto master. But if you get it working on all platforms, then it will be good to adapt the code. No, no, don't work on all platforms, HP-UX bug fixing is a waste of time. OK, you got it working on anything but HP-UX? No, I just want it on Ubuntu 16.04 and don't care about the rest, for now.

It is quite discouraging.

------------

> So, to make baby steps which are easier to integrate I would like to have here all platfroms that don't required special handling of

> pycryptodome==${PYCRYPTODOME_VERSION}upstream

> and this should be targeted to pycryptodome so cryptography changes should be left for a separate pr

> cryptography==${CRYPTOGRAPHY_VERSION}upstream

What is the issue with this little hack? It solves problems for which we have no alternate solutions. If you are opposing the versioning, let's fix that. I am not keen to step back from this work for no reason other than a naming issue.

Also, drive-by fixes are an issue now? It's a related fix that brings latest `cryptography` on both OS X and SLES 11.

**needs-review**

PR #112, opened 2018-07-24
[#5026] Unbreak HP-UX by not using CFFI / cryptography. Use cryptography wheel on SLES 11 / OS X, but not on macOS.

dumol:
Scope
=====

Currently, in `python-package` we use latest `cryptography` in HP-UX (with a small patch).

But server tests crash with it, and consequently we use an older packaget for HP-UX with many security updates missing.

Changes
=======

Stop building CFFI/cryptography on HP-UX and revert to pyOpenSSL 0.13.x, as with Solaris 10.

GMPY 1.x sources were imported, to be built separately in HP-UX, where there are issues with GMPY 2.x, and we can't use the wheel any more.

All of the work has been done in the `5017-hpux-cffi-crash` branch. Here I have just cherry-picked commits and changes from that branch and its parent, `4985-pycryptodome-instead-of-pycrypto`. 

**Drive-by fixes** (also cherry-picked from the latter branch above, have passed review in #110):
  * Stop using upstream `cryptography` wheel on macOS, where we want it linked to Homebrew's OpenSSL libs.
 * Use latest `cryptography` on SLES 11, OS X 10.8 (and possibly RHEL 5) through upstream wheels, that come with OpenSSL 1.1.0h. `stdlib ssl` is still linked to local 0.9.8, eg. https://chevah.com/buildbot/builders/python-package-sles-11/builds/44/steps/test/logs/stdio

reviewers: @adiroiban 

Please review changes in `chevah_build` and `python-modules/chevah-python-test/test_python_binary_dist.py`. Or altenatively the last two commits.

Run automated tests, eg. https://chevah.com/buildbot/builders/python-package-group-all/builds/28

Test server in HP-UX with a testing package from this branch, eg. https://chevah.com/buildbot/builders/server-hpux/builds/66


dumol:
As this is possibly the last PR before switching to `pycryptodome`, it might be worth considering updating PyCrypto from master where possible, by integrating the unfinished changes in #108.

**needs-review**

chevah_build
```diff
@@ -323,17 +323,13 @@ case $OS in
         export MAKE="make -P"
         export BUILD_LIBFFI="yes"
         export BUILD_ZLIB="yes"
+        # CFFI/cryptography build on HP-UX, but libffi breaks for CFFI. More at
```
adiroiban: Maybe
CFFI/cryptography can be built/compiled on HP-UX but is not usable.

chevah_build
```diff
@@ -342,10 +338,11 @@ case $OS in
         export CXX="clang++"
         export CFLAGS="$CFLAGS -mmacosx-version-min=10.8"
         export MACOSX_DEPLOYMENT_TARGET=10.8
-        # OS X 10.8 not supported by cryptography 1.9, so we use 1.8.2.
+        # OS X 10.8 not supported by cryptography 1.9, so you'll normally use
+        # 1.8.2, but cryptography 2.2.2 has a working wheel for MacOS 10.6+.
```
adiroiban: Mabye we should mention why the wheel works.

I guess that it works as comes with a static OpenSSL?

If we have static OpenSSL in OSX and RHEL5/SLES11

Once this is used in server we need to update the documentation at https://www.sftpplus.com/documentation/sftpplus/latest/guides/cryptography.html#openssl-versions-available-in-most-popular-operating-systems

Does this make sense?

chevah_build
```diff
@@ -386,8 +383,15 @@ case $OS in
         export BUILD_LIBEDIT="no"
         ;;
     rhel5|sles11)
-        # OS'es with OpenSSL 0.9.8 libs, older cryptography needed here.
-        PIP_LIBRARIES=$PIP_LIBRARIES_OPENSSL_098
+        # OS'es with OpenSSL 0.9.8 libs, older cryptography would be needed,
+        # but upstream provides a Linux wheel that works, at least on SLES 11.
```
adiroiban: it talks about "at least SLES11" but I see that this is also applied to RHEL5.

I think that we should not mention the SLES 11 but rather explain why the upstream wheel works.

adiroiban:
> it might be worth considering updating PyCrypto from master where possible, by integrating the unfinished changes in #108.

I think that we can drop the work for pycrypto. pycrytodome is the future and once it builds on HPUX we should put all our effort into pycrytodome

chevah_build
dumol: Maybe… But  my message also conveys that it's because of libffi, so I still prefer it.

chevah_build
dumol: True, those wheels for OSX and RHEL5/SLES11 come with static OpenSSL 1.1.0h, as mentioned in the PR. I will add a mention in the comments here as well.

Documentation should be updated, of course. It's already outdated for AIX and Solaris.

chevah_build
dumol: Yes, I can add here a message about the bundled OpenSSL as well.

I wrote _at least on SLES 11_ because that's what I tested. I haven't tested this on RHEL 5, so I'm not sure it will work. Hopefully we'll never need to find out. :-]

PR #115, opened 2018-09-04
[#4974] Test for outdated and insecure Python modules with pip and safety.

dumol:
Scope
=====

We might want to look into setting up something like ​​https://pyup.io/safety/

Is free for on-premise deployment.

Changes
=======

Use `pip list --outdated` to show old Python modules.

Use `safety` to check for known vulnerabilities in Python modules and error out if any found. 
But do not bundle `safety`, only install it with `pip` during test phase.

**Drive-by fixes**:
  * updated `pip` and `get-pip.py` to latest version: 18.0
  * also copy `paver.conf` alongside `paver.sh` for compat tests, to avoid incompatibilities
  * updated `pypiwin32` and renamed accordingly to upstream to `pywin32`, added test for it
  * updated `scandir` to 1.9.0 (this fixes testing in OpenBSD)
  * install some modules through `pip` even on Solaris 10 and HP-UX using `PIP_LIBRARIES_NO_CFFI`
  * adjusted list of required packages for SLES 12 to account for the SP3 upgrade.

reviewers: @adiroiban 

Please double-check changes in `chevah_build` and `python-modules/chevah-python-test/test_python_binary_dist.py`

Run the automated tests and check for the output of our tests.
`pip` should list **all** outdated packages, eg. pyOpenSSL everywhere, as 18.0.0 was released and we are at most at 17.5.0.
`safety` should list **all** packages with known vulnerabilities.

Make sure the test phase errors out if any vulnerabilities are found.
Note that for some reason `safety` doesn't error on Windows (and only on Windows), eg. https://chevah.com/buildbot/builders/python-package-win-xp-dev/builds/62/steps/test/logs/stdio. Reported upstream at https://github.com/pyupio/safety/issues/153.


chevah_build
```diff
@@ -64,6 +65,7 @@ EXTRA_LIBRARIES_NO_CFFI="\
     python-modules/pycryptodome-${PYCRYPTODOME_VERSION} \
     python-modules/python-setproctitle-1.1.8.dev0 \
     python-modules/python-scandir-${SCANDIR_VERSION} \
+    python-modules/python-safety-${SAFETY_VERSION} \
```
adiroiban: Do we need it from python-modules?
I think that we can get it via pip  from pypi.chevah.com

in python-modules we only used to have the code this is CPU/OS dependent

safety is an universal wheel..


Note that I have already start using safety for chevah/server see https://github.com/chevah/server/blob/master/pavement.py#L1861

adiroiban:
I can take a look and see why safety is not failing on Windows.

-------

We don't always need/want the latest packages.
As long as there are no bugs and no security features we can work with older version.

But we can have `pip list --outdated` as an informational test as is always good to have the latest version.
Some packages are only used for development (ex click) and should be not present in the public product release.

Thanks!

chevah_build
dumol: That's what I tried first, but it didn't work for all OS'es. I'll look closer when I'll come back to this PR.

dumol: OK, so I took a closer look… The way things currently work in our build system, for platforms not built around cffi, `PIP_LIBRARIES` is uninitialized and these modules (PyOpenSSL, PyCryptodome etc.) are collected in `EXTRA_LIBRARIES` and built separately through `initialize_python_module()` using `setup.py`. Currently supported platforms in this situation: Solaris 10 and HP-UX.

What we can do here, I guess, is to get through `pip` the modules that would work, like `scandir` and `safety`, using something like `PIP_LIBRARIES_NO_CFFI` for uniformity. I've tried something like that for Solaris 10 and it worked: https://chevah.com/buildbot/builders/python-package-solaris-10/builds/72

Do you think this is worth implementing? The end result is identical, as far as I can tell.

dumol:
> I can take a look and see why safety is not failing on Windows.

I suspected a shell issue, but the error code from `safety` when executed on Windows is `0` even when vulnerabilities are found, eg. https://chevah.com/buildbot/builders/python-package-win-xp-dev/builds/69

So this seems like an upstream bug…

dumol:
Have reported the `safety` issue on Windows at https://github.com/pyupio/safety/issues/153. Let's see what upstream thinks about it…

chevah_build
adiroiban: we need to create 2 types of  Python package/modules, similar to what we have for chevha/server.

Runtime:
* scandir
* cryptography
* pyopenssl

Build time / development helpers:
* brink
* safety
* ... etc

------

I don't want to have `safety` part of the python-package distribution.

Since `safety` has no CPU/OS depended part, we can install it separately in each project and be able to update `safety` without updating python-package.

So safety should be part of the pavement.py development script or implemented in a similar way

dumol: Yes, it's better to not bundle `safety`. But I've already implemented `PIP_LIBRARIES_NO_CFFI` for Solaris 10 and HP-UX, hope you don't mind.

So `safety` is installed now only during the test phase, when actually needed. I'm not sure about `pavement.py`, that is not used when building or testing locally. What do you have in mind?

dumol:
**needs-review**

chevah_build
```diff
@@ -89,10 +86,15 @@ PIP_LIBRARIES_OPENSSL_098="\
     pyOpenSSL==16.0.0 \
     scandir==${SCANDIR_VERSION} \
     "
+# Python modules installed through pip on systems not build around cffi.
```
adiroiban: Maybe use this comment.

Python modules resulting in OS/CPU specific binaries, but which don't depend on cffi.

---------

The idea is that we have cffi working, with the exception of cryptography, all the other python modules no longer create OS/CPU specific binaries.

dumol:
Thanks for the review!

> Not sure why scandir is not reported as vulnerable.

Why do you think scandir 0.7 was vulnerable? Have updated it as a drive-by fix. Actually wasn't intending it, but I removed scandir sources from `python-modules/` after implementing PIP_LIBRARIES_NO_CFFI, as they were no longer required for building scandir, even on Solaris 10 and HP-UX. But then I realized we actually need most files in there for running the scandir tests, so I had to put them back. And at that moment I thought it would be better to put latest stable version, as the OpenBSD issue was fixed upstream.

chevah_build
dumol: I don't know… We don't split them on this criteria for other OS'es.

adiroiban:
> Why do you think scandir 0.7 was vulnerable?

Sorry about that. I saw that scandir was a drive-by but it did not "clicked" that it is safe.

All good.

dumol:
Outdated packages checked only on our server. See above comment for details.

**needs-changes**

dumol:
Please re-review. Thoroughly, as there were some important changes and PR's text has important additions. Some are not really needed (such as the `get-pip.py`, `pip` and `pypiwin32` -> `pywin32` updates), so I can remove them if you think they might get in the way. 

Although `pypiwin32` -> `pywin32` looks like it could be a security update, judging by the changelog at https://github.com/mhammond/pywin32/blob/master/CHANGES.txt (look for _leak_ and _crash_ until you hit _Since build 218_).

On a related note to PIP_LIBRARIES_NO_CFFI and building more stuff through pip, I couldn't find a switch for pip to show more output for the compilation in background. I was looking through https://pip.pypa.io/en/stable/reference/pip_install/.  Any other ideas? Thanks!

**needs-review**

adiroiban:
> Some new updates are not really needed (such as the get-pip.py, pip and pypiwin32 -> pywin32 updates), so I can remove them if you think they might get in the way.

Yes. I think that get-pip.py can be removed as latest python comes with pip... I think that only on Solaris  SPARC we still need it, as the old Python version from Solaris has no pip.

-----------

> Although pypiwin32 -> pywin32 looks like it could be a security update,

Yes. We should stop using pypiwin32 and just use pywin32.
For some time, the upstream dev had not time to publish binary wheels for pywin32 so someone else create pypiwin32 and uploaded the wheels.
In this we can use pywin32 without compiling it on Windows.
So pypiwin32 is obsolete and we should switch to pywin32.

--------

> On a related note to PIP_LIBRARIES_NO_CFFI and building more stuff through pip, I couldn't find a switch for pip to show more output for the compilation in background. I was looking through https://pip.pypa.io/en/stable/reference/pip_install/. Any other ideas? Thanks!

Which output is missing / What information do you want to see for pip install?

Does `pip install -v PKG1 PKG2` give what you want?

Btw. I think that we should always using `pip install -v` to see whether a new binary was created or a cached wheel was used.

Modern pip versions are keeping a cache of all installed packages as wheels.
If a package does not has a wheel on PYPI, pip will create one and stored on the local cache and use it when needed, without recompiling.

--------

I will look at the changes a bit later. I just wanted to answer the questions

dumol:
Thanks for the answers and the review!

> I think that only on Solaris SPARC we still need it, as the old Python version from Solaris has no pip.

To be precise, Solaris 10 on both SPARC and X64 is stuck with Python 2.7.8.

> Does pip install -v PKG1 PKG2 give what you want?

The output is more verbose, for sure. But I still can't see the compilation phase for `pycryptodome`. Sometimes there is interesting output even when building succeeds, as it was the case for the recent issue with RHEL 6. I suppose `-v` is a general switch to `pip`, so I see it can be repeated 3 times for increased verbosity. However, neither `-vvv` or `-v -v -v` help much in this regard.

So, to be honest, I would prefer building `pycryptodome` from local sources, as we do it for Solaris 10 and HP-UX. What do you think?

> idna, enum32, ipaddress are pulled by cryptography... not sure why we don't get the latest version for them, but as long as there is no know security we should be fine.

Because the packages are pulled from our server only?

adiroiban:
> Because the packages are pulled from our server only?

True :) Thanks

> So, to be honest, I would prefer building pycryptodome from local sources, as we do it for Solaris 10 and HP-UX. What do you think?

Sure. No problem. THanks!

dumol:
> > So, to be honest, I would prefer building pycryptodome from local sources, as we do it for Solaris 10 and HP-UX. What do you think?

> Sure. No problem. THanks!

OK, to be tried in another PR as a drive-by, if feasible. This one has been going for too long.

PR #116, opened 2018-09-05
[#5057] Security updates for pycryptodome and cryptography.

dumol:
Scope
=====

As suggested by `safety` while working in #115:

-> pycryptodome, installed 3.6.5, affected <3.6.6, id 36384
pycryptodome before 3.6.6 has a vulnerability on AESNI ECB with payloads smaller than 16 bytes.

--
-> cryptography, installed 2.2.2, affected >=1.9.0,<2.3, id 36351
python-cryptography versions >=1.9.0 and <2.3 did not enforce a minimum tag length for finalize_with_tag API. If a user did not validate the input length prior to passing it to finalize_with_tag an attacker could craft an invalid payload with a shortened tag (e.g. 1 byte) such that they would have a 1 in 256 chance of passing the MAC check. GCM tag forgeries can cause key leakage.

Changes
=======

Updated PyCryptodome to 3.6.6 and cryptography to 2.3.1 in `chevah_build`, `python-modules/` sub-dir and on https://pypi.chevah.com.

**Drive-by fixes**:
  * For OpenBSD there is no longer a need for patching `cryptography`.
  * Identifying what `cryptography` version to use in different Solaris 11.x versions works in a more simple way now.

reviewers: @adiroiban 

Review the changes in `chevah_build` or the latest two commits.
Run the automated tests.


dumol:
**needs-review**

PR #117, opened 2018-09-13
[#5023] Debian-specific packages.

dumol:
Scope
=====

As discussed in Trac ticket #4986 and on private IRC, we should have specific Python packages for Debian (both x86 and x64).

The updated `paver.sh` should be integrated into server.

Changes
=======

Treat Debian as supported Linux distro.

reviewers: **WIP**

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -367,11 +367,6 @@ case $OS in
         export CXX="clang++"
         export BUILD_LIBFFI="yes"
         ;;
-    linux*)
-        # We don't compile libedit for generic Linux builds because it links
-        # to local ncurses libs and the result is not very portable.
-        export BUILD_LIBEDIT="no"
-        ;;
     sles11sm)
         # Use PyCryptodome's wheel, as there are issues not solved upstream yet.
         PIP_LIBRARIES="\
@@ -457,7 +452,7 @@ check_dependencies() {
 
     case $OS in
         # Debian-derived distros are similar in this regard.
-        ubuntu*|raspbian*)
+        ubuntu*|debian*|raspbian*)
             packages=$DEBIAN_PKGS
             check_command='dpkg --status'
             ;;
--- a/paver.sh
+++ b/paver.sh
@@ -640,6 +640,10 @@ detect_os() {
                     # Arch Linux is a rolling distro, no version info available.
                     OS="archlinux"
                     ;;
+                *)
+                    echo 'Unsupported Linux distribution:' $ID
+                    exit 15
+                    ;;
             esac
         fi
     elif [ "${OS}" = "darwin" ]; then
--- a/paver.sh
+++ b/paver.sh
@@ -618,6 +618,12 @@ detect_os() {
                         echo "Unsupported Ubuntu, using generic Linux binaries!"
                     fi
                     ;;
+                "debian")
+                    os_version_raw="$VERSION_ID"
+                    check_os_version "$distro_fancy_name" 7 \
+                        "$os_version_raw" os_version_chevah
+                    OS="debian${os_version_chevah}"
+                    ;;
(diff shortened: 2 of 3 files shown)


dumol:
Compat tests are currently failing. I think `compat` master is a bit left behind. Can you please double-check? Thanks!

**needs-review**

chevah_build
```diff
@@ -880,11 +875,11 @@ command_test_compat() {
     execute rm -rf compat
     execute git clone https://github.com/chevah/compat.git --depth=1 -b master
     execute pushd compat
+    # Copy over the latest paver stuff, as some changes might require it.
+    execute cp ../../paver.{conf,sh} ./
```
adiroiban: copying paver.conf is not quite ok... but for now I don't think there are issues.

compat, might have different base requirement, especially brink. so using a different version of brink might break the compat buiild system.

I think that we should copy just paver.sh and overwrite only PYTHON_CONFIGURATIO

dumol:
needs-changes

dumol:
> Once this is in production, please consider updating chevah/server to use the new package names.

Can't do this until the transition to PyCryptodome is completed in https://trac.chevah.com/ticket/5034.

Also, even with not overwriting `paver.conf`, compat tests fail everywhere. The reason they worked in latest revisions for master (in `python-package`) is because `paver.conf` was overwritten after the current Python version was configured for compat tests. Therefore, `compat` tests were run for an older Python version.

Can you please look at updating compat master? The error everywhere is:

```
Installing base requirements: chevah-brink==0.69.1 paver==1.2.4.
  File "<string>", line 1, in <module>
ImportError: No module named paver.tasks
```

I wouldn't want to commit this to master with broken compat tests everywhere. Thanks!

**needs-review**

dumol:
`PYTHON_CONFIGURATION='default@2.7.15.1afde4e1:solaris10@2.7.8.1afde4e1'` is the latest revision with PyCrypto and the one to hit server master really soon now in https://github.com/chevah/server/pull/1443

adiroiban:
Thanks.
Latest python looks ok on my local system.

The changes from this PR looks good. .. not sure why compat tests are no longer working.

The compat tests need to run in a similar way to buildbot. That is run `paver deps` before `test_ci`.

I am testing this on my local computer...

adiroiban:
Ok. So the issue is the new `pip` version

```
    Skipping link http://pypi.chevah.com/simple/chevah-brink/chevah-brink-0.58.0-py2.py3-none-any.whl (from http://pypi.chevah.com/simple/chevah-brink/); wrong project name (not chevah-brink)
```

Somehow it does not work with our packages.

I think that latest pip wants `chevah_brink-0.58.0-py2.py3-none-any.whl` . That is use underscore in project name.

My suggestion is to use an older pip version for now... while we look at how to update the package names

adiroiban:
I think that the current trick with static files will not do the trick, so we might look at installing https://github.com/pypiserver/pypiserver on the public server and no longer have pypi on deag

adiroiban:
the issue is that we need to force using older pip.
I have pushed a commit.

Not sure how you want to handle it

I guess that in order to close this fast we should use pip 9.0.0.

Then, in a separate task we can look at upgrading our pypi servers and then start using latest pip.
As part of the separate task, we should test chevah/server branch `series-2` that it can still use the pypi server.. so that we don't end up breaking SFTPPlus V2 dev

needs-changes

dumol:
Thank you for the fix. I've set `9.0.3`, that was used extensively, to be used in the same way as before. We don't need latest `pip` for now, so I'm fine with the current situation, no need to complicate things unnecessarily.

**needs-review**

adiroiban:
thanks. hope all tests will be green
changes-approved

PR #119, opened 2018-10-25
[#5103] Build libffi on SLES.

dumol:
Scope
=====

We have inadvertently built our Python packages against a `libffi6` package from the _monitoring_ OpenSUSE repo for SLES, used for installing Zabbix.

Changes
=======

Just don't use any local `libffi` on SLES, as even the `libffi-devel` packages from the SDKs DVDs or repos try to replace the `libffi4` package available to be installed in a default setup with `libffi43` packages that come with no vendor support.

We have a similar situation with AIX / HP-UX / the BSDs, where we use our own `libffi`.

To be extra safe, have removed `libffi-devel and libffi6` from SLES minions through Salt.

**Drive-by fixes**:
  * updated `paver.sh` with latest Ubuntu-related improvements from https://github.com/chevah/salt-master/pull/391
  * use stderr when erroring out from `chevah_build`
  * removed unused `libffi-3.3-rc0` sources.

How to try and test the changes
===============================

reviewers: @adiroiban 

Please review changes.

Run the automated tests.

PR #121, opened 2018-11-12
[#5124] Amazon Linux support.

dumol:
Scope
=====

Support for Amazon Linux 2.  Using ​https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/amazon-linux-2-virtual-machine.html.

Changes
=======

Recognize `amzn` Linux distribution type.

Support Amazon Linux version 2 and newer.

**Drive-by fix**:
  * nicer exit message for unsupported Linux distros with `/etc/os-release` files.

reviewers: @adiroiban 

Please review the simple changes.

Manually test on one of the new `amazon` slaves on Fiser and Ghindari, not yet integrated in our Buildbot setup. All tests pass with no changes other than the dep checks.

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -464,7 +464,7 @@ check_dependencies() {
             packages=$DEBIAN_PKGS
             check_command='dpkg --status'
             ;;
-        rhel*)
+        rhel*|amazon*)
             packages=$RHEL_PKGS
             check_command='rpm --query'
             ;;
--- a/paver.sh
+++ b/paver.sh
@@ -641,7 +641,7 @@ detect_os() {
                     OS="archlinux"
                     ;;
                 *)
-                    echo "Unsupported Linux distribution type: $linux_distro."
+                    echo "Unsupported Linux distribution: $distro_fancy_name."
                     exit 15
                     ;;
             esac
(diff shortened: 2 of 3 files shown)


dumol:
> Thanks. Please publish it to production.
> […]
> Once this is merged, please create a ticket in chevah/server in which we will add this new Linux version.

Thanks! This was the support for master. To upload binaries to production ASAP, I'll have to apply these changes to a branch with PyCrypto, eg. one derived from #119. With the uploaded binaries I'll enable the build slaves for Amazon Linux 2 and start working on server.

> I have assisted a client running SFTPPlus rhel 7.4 in Amazon Linux 2 and it worked.
> 
> It was strange, as systemd was not there.

Hhmmm, judging by the deps, some things should not work, even on the older 7.x package. But maybe the problematic deps are for modules not used by SFTPPlus.

> I went for sysV script.
> 
> For the sysV script we should also need a variant in which the process is not started as root, but directly as the "sftpplus" user.

I don't see/understand the reason for supporting running SysV init scripts outside a SysV-compatible init system. You can run SFTPPlus as a dedicated user through SysV, but all SysV init scripts are run by root. If you don't like that, you can use a script to run your daemon outside the SysV system, but it's not going to be a SysV init script. Do we want such a script, in addition to the SysV init one?

PR #123, opened 2018-11-15
[#5115] Bring back PyCrypto alongside PyCryptodome.

dumol:
Scope
=====

Get PyCrypto back.

Ideally with PyCryptodome built as `pycryptodomex`, to not lose all the work invested in getting that working on everything but HP-UX.

More on this in internal Trac ticket #5040.

Changes
=======

Use PyCryptodome 3.6.6 as PyCryptodomex.

Build PyCrypto 2.6.1 again, but this time always from `EXTRA_LIBRARIES`. Imported pristine sources + patches separated in a corresponding sub-dir.

But in the end, for Windows I had to pack the patched sources and upload them to PyPI server anyway, as `pycrypto-2.6.1chevah1`… ¯\\\_(ツ)\_/¯

Apply all Ubuntu 16.04 patches to PyCrypto's sources  + our own patch for AIX. This way we patch two security vulnerabilities for PyCrypto 2.6.1: CVE-2013-7459 and CVE-2018-6594.

**Drive-by changes**:
 * added a way to exclude false positives from `safety` checks for pyOpenSSL 0.13.1 and patched PyCrypto 2.6.1
 * added a mechanism for keeping changes to stuff in `python-modules/` in separate patches, eg. for `python-modules/pycrypto-2.6.1`, patches are applied from `python-modules/pycrypto-2.6.1-patches`
 * build `zlib` on macOS to account for missing headers in 10.14
 * build `libffi` on Solaris 11 to account for issues with multiple `libffi` libs in 11.4
 * build `libedit` and `readline` modules on Solaris 11 SPARC again
 * identify Solaris 11 older than 11.4 as `solaris-112`, following the scheme for older Solaris 10
 * on Solaris 11.2, get back to pyOpenSSL 0.13.1, as 0.17.0 is vulnerable, and the patch for it requires a version of cryptography that doesn't run on Solaris 11.2.
 * use required latest `cryptography` (and pyOpenSSL 0.17.5) in Solaris 11.4 (there's no longer a need for patched `chevah` versions and our fork in GitHub)
 * also updated `cryptography` to latest version on all OS'es using it, thus also OpenSSL version for Windows / SLES 11 / OS X
 * while at it, also updated pyOpenSSL to latest version: 18.0.0.

reviewers: @adiroiban 

Please review changes.

For my changes: `git diff master paver.sh chevah_build functions.sh python-modules/chevah-python-test/test_python_binary_dist.py src/`

For Ubuntu patches + our patch for PyCrypto: `git diff master --text python-modules/pycrypto-2.6.1-patches/`

Everything else should be just source renames and imports.

Run the automated tests.


dumol:
**needs-review**

adiroiban:
I think that for Pycrypto we should try to pathc.

For pyopenssl, I am not sure which vulnerability is reported. I am looking at https://snyk.io/vuln/pip:pyopenssl

But the null byte handler should have been fixed in 0.13.1.

For the other error, I don't think that we are affected - https://people.canonical.com/~ubuntu-security/cve/2018/CVE-2018-1000807.html

0.13.1 don't have the new code which is vulnerable.


So I think that 0.13.1 is fine.

Is just that safety should tag the CVE as > 0.13.1  

--------

but everywhere else, we need to upgrade pytonnssl to 17.5.0... or better 18.0.0 :).

functions.sh
```diff
@@ -306,3 +306,21 @@ aix_ld_hack() {
             ;;
     esac
 }
+
+#
+# Safety DB IDs to be ignored when using pyOpenSSL older than 17.5.0.
+#
+add_ignored_safety_ids_for_pyopenssl_pre17_5_0() {
```
adiroiban: I don't know what to say....

I would not ignore this and just upgrade pyopenssl.

These CVEs don't affect 0.13.1

dumol:
So there's only one place where we use a pyOpenSSL version newer than 0.13.1, but older than 17.5.0, and that is Solaris 11. In theory this could be solved by upgrading our Solaris 11 installations to 11.3 SRU 19 or later. Which in practice means reinstallation everywhere, as detailed in https://trac.chevah.com/ticket/4918.

Even if we fix this for Solaris 11, there will still be a need to ignore these IDs for pyOpenSSL 0.13.1 in Solaris 10 and HP-UX (possibly also on SLES 10, but nobody cares). I could rename the function from `add_ignored_safety_ids_for_pyopenssl_pre17_5_0` to `add_ignored_safety_ids_for_pyopenssl_13_1` to account for that. Alternatively, we could also file an issue with `safety` upstream, but in my experience they ignore their free users.

On the other OS'es I could also upgrade pyOpenSSL to latest version (currently 18.0.0), but that is not a security issue, and wouldn't change anything for this. Would prefer a separate branch/PR.

Would the above be acceptable?

adiroiban:
I think that we should drop support for Solaris 11.
I don't think that we have customers on Solaris 11. I know that we have only for Solaris 10.
@tim441 can confirm if we sold any Solaris 11 license.

I would say that we should drop support for Solaris 11 as is too much work to keep it secure.

--------

>  Alternatively, we could also file an issue with safety upstream, but in my experience they ignore their free users.

I think that we should report this to the `safety` project, so that they can update their DB to only raise these warnings if pyopenssl 14.0 or newer is found.

If they will ignore it, then this is a bad sign, and yes. we should create a function lik

`add_ignored_safety_pyopenssl_false_positives`

But I hope that they can update their db based on our feedback.

-------

Yes pyopenssl for the other OS should be upgraded in another branch, but it can wait. Maybe we will upgrade to 19.0.0 :)

Thanks!

tim441:
I am unaware of any Solaris 11 users and agree it makes sense to drop this

adiroiban:
I now see that the free db is updated once per month.
To run it on your own buildbot and get real time updated you need to pay $120 per month.

There is a cheaper version, but in this case they will check your code on GitHub and report errors...so you don't have full control

dumol:
I remembered this message on support, might be relevant:

On 05/09/18 13:51, Lowdon Peter DWP TECHNOLOGY wrote:
> Hi Adi,
> 
> This is the breakdown of the estate
> 
> Tiny fraction - Solaris 11.3 (SPARC)
> 
> Single server - Solaris 10 update4 (SPARC)
> 
> Vast Majority - Solaris 10 update11 (SPARC)
> 
> Handful of servers - Solaris 9 (SPARC)
> 
> Hope this helps.

I do have a mild interest in keeping Solaris 11 supported, so I will spin up a Solaris 11.4 x86 VM for `python-package` in my free time if Solaris 11 support is to be completely dropped. As I did for AIX 6.1 on Foi.

Note that DWP is using Solaris 11.3 on SPARC, so it's a slightly different matter. On a second thought, I think we could alternatively spin up kernel zones with updated Solaris 11.3 and/or 11.4 on SPARC using Solaris 11.2 on Jacu. Not sure how time-consuming that would be in comparison to simply reinstalling a newer Solaris version on Jacu.

dumol:
For the record, I reported upstream the `safety` issue with pyOpenSSL 0.13.1 at https://github.com/pyupio/safety/issues/174.

Also took the time to update the spreadsheet for _SFTPPlus external dependencies and associated vulnerabilities_ at https://docs.google.com/spreadsheets/d/12N64qFknk91f0YzEFEZ62jJ8VHfhrX2OT0nNTCn9bvY/.

dumol:
**needs-review**

adiroiban:
Looks good.
But now on SLES-11 SM we have  1.0.1, while on normal we have 1.1.0... maybe we should drop SLES 11-SM and just keep the SLES-11 with embedded OpenSSL

adiroiban:
yes... not sure what the distros will do about 1.0.1... will see. thanks!

PR #125, opened 2019-01-09
[#5180] Build psutil on supported platforms.

dumol:
Scope
=====

Adding `psutil` support.

Changes
=======

Build `psutil` everywhere except:

  * Windows, where we use the upstream wheel
  * HP-UX, for which there is no support from uppstream
  * AIX 5.3, where the build fails.

**Drive-by fixes**:
 * updated deps for Alpine 3.7 and Debian 8
 * build Debian packages as portable as possible, there might be of some use on other unsupported Linux platforms. This means no `readline` module on Debian.

reviewers: @adiroiban 

Please review changes.

Run the automated tests.

### DIFF
--- a/python-modules/chevah-python-test/test_python_binary_dist.py
+++ b/python-modules/chevah-python-test/test_python_binary_dist.py
@@ -854,7 +854,9 @@ def main():
             exit_code = 18
 
     # Some OS'es are not supported by upstream psutil.
-    # Windows is special. We build on XP, but XP/2003 not supported any more.
+	# FIXME:5185:
+    # Windows is special. We build python-package on XP, with psutil as wheel,
+	# but XP/2003 not supported any more upstream, so we can't test for it.
     if not chevah_os in [ 'aix53', 'hpux1131', 'windows' ]:
         try:
             import psutil
--- a/python-modules/chevah-python-test/test_python_binary_dist.py
+++ b/python-modules/chevah-python-test/test_python_binary_dist.py
@@ -750,13 +750,6 @@ def main():
         exit_code = 17
 
     try:
-        import psutil
-        print 'psutil %s' % (psutil.__version__,)
-    except:
-        sys.stderr.write('"psutil" missing.\n')
-        exit_code = 23
-
-    try:
         import gmpy2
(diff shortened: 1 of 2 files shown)


dumol:
**needs-review**

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -844,6 +853,15 @@ def main():
             sys.stderr.write('"_scandir" missing.\n')
             exit_code = 18
 
+    # Some OS'es are not supported by upstream psutil.
+    # Windows is special. We build on XP, but XP/2003 not supported any more.
+    if not chevah_os in [ 'aix53', 'hpux1131', 'windows' ]:
```
adiroiban: I think that windows should not be included here.... instead we should have a separate test for windows in which we take into consideration the windows version/

You can use this code to get the windows version in the format nt-6.1 nt-6.2

```
    if os.name == 'nt':
        parts = platform.version().split('.')
        return 'nt-%s.%s' % (parts[0], parts[1])
```

and then ignore import error only on some Windows version....

with that in mind, we need a separate python-package slave for  Windows... maybe based on Windows-2016 so that we can run this test on Windows.

adiroiban: Also, a better test would be to try to import psutil and then ignore only ImportError... in this way we can cover the case in which psutil might be there but raises a different error

dumol:
With these changes we do have `psutil` in the Windows package, as a wheel. It's just not tested, because it bails out on import with an error about the Windows version on our XP slaves.

I agree we should try using a newer Windows version for building `python-package`. But this looks like a major undertaking, lots of things are different in the Windows build. And we don't have AdiL with us any more…

Then we can fix this minor issue with checking for `psutil` on Windows XP. But for now it doesn't really make sense to write code that it's not run.

**needs-review**

python-modules/chevah-python-test/test_python_binary_dist.py
dumol: Thanks for the suggestions… I've added a pointer to these comments at https://trac.chevah.com/ticket/5185#comment:1.

PR #126, opened 2019-02-21
[#5203] Cython addition.

dumol:
Scope
=====

see if it is easy to get cython for our Linux and Windows python-package version

MacOS is optional.

Changes
=======

This was completely problem-free for all the platforms.

On Windows I have used the upstream wheel, which according to their documentation requires using Visual C on Windows.

**Drive-by fixes**:
  * updated SQLite for latest security issues: https://blade.tencent.com/magellan/index_en.html
  * detect MSYS env under Windows.

reviewers: @adiroiban 

Check changes, eg. with:
```
git diff master paver.conf chevah_build python-modules/chevah-python-test/ paver.sh
```

Run the automated tests.


dumol:
If you want other updates besides the SQLite one, please let me know.

I'm only aware of a moderate issue with OpenSSL 1.0.2, to be fixed on Feb 26 by upstream. This will probably be included in Python 2.7.16, to be released on March 2. Only affects Windows.

**needs-review**

adiroiban:
Thanks. Great work.
I need to write an example of cython and make sure that cython is usable... just importing cython is a first step, but it will not guarantee that it can compile .pyx files.

I will write a demo .pyx file and update the pytho module test code to compile a pyx file and then use it in the test

dumol:
Thanks, that should also work on Windows, our current XP slaves for `python-package` have Visual C installed.

True, there is a Trac ticket for using newer Windows build slaves for this (#5192). And I'm currently thinking of jumping straight to Windows 2019, and leverage Microsoft's support for backward compatibility when building on them.

adiroiban:
> jumping straight to Windows 2019

Windows 2019 looks good and if you add a Windows 2019 slave, please also install Python 2.7 from upstream in C:/Python27 and then install https://www.microsoft.com/EN-US/DOWNLOAD/DETAILS.ASPX?ID=44266

----------

I saw that MS did an ugly thing with Windows 10... they released a major upgrade but still kept the version number as 10.

Is good that Windows Server version was updated.

It was confusing as I had Windows 10 installed on my laptop and there was an app advertised as only working on Windows 10 and installable from Windows Store... but when I went on Windows store it told me that the software is not available for me... without any hint that I should update.
Bad user experience.

But we are not using any of the Windows 10 features... we are not even using Windows 8/2012 features.

dumol:
@adiroiban, any news on this front? 

Would like to get Python 2.7.16 (has security fixes) and cryptography 2.6 (with updated OpenSSL wheel on Windows) into `python-package`.

Should I do it in this branch or separately?

dumol:
Thanks, will merge this into master then. And open a dedicate ticket for updating Python and cryptography…

PR #127, opened 2019-03-14
[#5225] Updates for Python,  cryptography, etc.

dumol:
Scope
=====

`python-package` with latest Python and `cryptography` for updated OpenSSL wheels.

Also check for other possible security updates while at it.

Changes
=======

Mostly uneventful updates:

  * Python 2.7.15 -> 2.7.16 (OpenSSL wheel on Windows updated from 1.0.2o to 1.0.2q)
  * cryptography 2.4.2 -> 2.6.1 (OpenSSL wheel on Win / OS X / SLES 11 updated from 1.1.0j to 1.1.1b)
  * pyOpenSSL 0.18.0 ->  0.19.0 (for compatibility with OpenSSL 1.1.1, as documented upstream)
 * cffi 1.11.5 -> 1.12.2 (fixes some minor bugs)
 * scandir 1.9.0 -> 1.10.0 (not really needed after all).

reviewers: @adiroiban 

To review changes: ``git diff master chevah_build python-modules/chevah-python-test/``

Latest `cryptography` fails on XP, but do we care? It works fine for Windows Server 2019, which should replace the XP slaves for `python-package` once `compat` tests are fixed in https://github.com/chevah/compat/pull/94.


dumol:
Compat tests fail for salted POSIX systems, to be fixed in https://github.com/chevah/compat/pull/94.

Same for `win-2019` compat tests, which are the only thing currently preventing us to finally replace `win-xp` with `win-2019`.

**needs-review**

chevah_build
```diff
@@ -933,7 +932,6 @@ command_test() {
             # available, but they need to be requested for the tests to pass.
             export LC_ALL=en_US.UTF-8
             execute $LOCAL_PYTHON_BINARY ${SCANDIR_FOLDER}/test/run_tests.py
-            export LC_ALL=C
```
adiroiban: why this change?
The idea is that all tests should run with `LC_ALL=C`... in this way, Python does not trigger the implicit Unicode conversion.

We run scandir tests with non LC_ALL=C as some tests are poorly written

dumol: Sorry, my bad! I will revert this…

Had some issues with testing `scandir` on Solaris 10, so I tried something here, intending to also update `scandir`. But things improved by themselves somehow, so there wasn't any need for changes or the update.

PR #131, opened 2019-05-08
[#5265] OpenBSD 6.5 cryptography build fix.

dumol:
Scope
=====

Cryptography fails to build on OpenBSD 6.5.

High prio, because I have upgraded to 6.5 and can't use paver stuff locally.

Changes
=======

Used the  OpenBSD -current  patches for `cryptography` 2.6.1 to build a special tarball to be used on OpenBSD 6.5

**Drive-by changes**:
  * updated paver stuff.

reviewers: **WIP**

Please review changes.

Run the tests.

### DIFF
--- a/chevah_build
+++ b/chevah_build
@@ -402,12 +402,7 @@ case $OS in
         # Beware that zlib headers are not present either on macOS 10.14.
         # We don't care here because we build on 10.13.
         ;;
-    freebsd*)
-        export CC="clang"
-        export CXX="clang++"
-        export BUILD_LIBFFI="yes"
-        ;;
-    openbsd*)
+    freebsd*|openbsd*)
         export CC="clang"
         export CXX="clang++"
         export BUILD_LIBFFI="yes"
--- a/chevah_build
+++ b/chevah_build
@@ -402,10 +402,28 @@ case $OS in
         # Beware that zlib headers are not present either on macOS 10.14.
         # We don't care here because we build on 10.13.
         ;;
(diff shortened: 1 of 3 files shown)


dumol:
Minimal changes in this PR, mostly for my OpenBSD desktop.

But I wonder if this would be a good time to enable optimizations when building Python, as advised during build:
```
If you want a release build with all optimizations active (LTO, PGO, etc), 
please run ./configure --enable-optimizations
```

**needs-review**

chevah_build
```diff
@@ -406,6 +406,19 @@ case $OS in
         export CC="clang"
         export CXX="clang++"
         export BUILD_LIBFFI="yes"
+        if [ "$OS" = "openbsd65" ]; then
```
adiroiban: this is ok, but I think that we should have support for single openBSD... and that should match the one used on your side.

So, this PR should also remove support for openbsd64

dumol: I agree with having a single release supported and tested for some OS'es, but I always keep in mind that you never know who is going to try this public repo. 

So, if easily achievable, I try to not break older versions, especially if still supported upstream.

dumol:
> 
> I think that we can enable optimizations for OpenBSD and Debian x86.. these are some version that we test but don't distribute to customers.
> 
> Also, we can see how much slower the optimized compilation end up.
> 
> and I think that is better to start with enabling one optimization at a time, rather than enabling all :)

OK, thanks! Will do that here as a drive-by before merging.

PR #136, opened 2019-10-17
[#5328] Improved support for macOS 10.14+, RHEL 8, OpenBSD 6.6, AIX.

dumol:
Scope
=====

Our current macOS Python package runs on macOS 10.14 and 10.15 too. But it would be nice to have `python-package` build on 10.15 with the newest compiler, while being tested on 10.13 with the stable build slaves for macOS. Akin to what we do in Windows, where we build on 2019, but test on 2012 and 2016 too…

Maybe switch from using LibreSSL 2.2.7 (default "OpenSSL" libs on 10.13) to using 2.6.x (default "OpenSSL" libs on 10.14, but also used for OpenSSH in 10.13). Although both versions are present in macOS 10.15, the 2.6.x LibreSSL libs would probably be supported for one more year with security updates by Apple.

Changes
=======

Copy over zlib headers, as we do for LibreSSL when building on macOS, because they are missing on 10.14+.

Have not switched LibreSSL version because version 2.2.7 is actually linked by default on both 10.13 and 10.14. OpenSSH doesn't seem to be linked to libs from `/usr/lib`. And `openssl` binary is not linked to default libs on 10.14. More on this in the associated Trac ticket.

Was able to use LibreSSL 2.2.7 when building on macOS 10.14 too, when linking 2.2.7 libs as default in `$INSTALL_FOLDER/lib`, which is prepended with `-L` in `LDFLAGS`. For reference the needed diff:
```
diff --git a/chevah_build b/chevah_build
index 5149fb07..c16fb8ff 100755
--- a/chevah_build
+++ b/chevah_build
@@ -676,6 +676,13 @@ command_build() {
                 $INSTALL_FOLDER/include/
             echo "Updating CFLAGS accordingly!"
             export CFLAGS="$CFLAGS -I${INSTALL_FOLDER}/include"
+            # Hack to link to LibreSSL 2.2.x libs on every macOS version.
+            execute pushd ${INSTALL_FOLDER}/lib/
+                execute ln -s /usr/lib/libssl.35.dylib
+                execute ln -s libssl.35.dylib libssl.dylib
+                execute ln -s /usr/lib/libcrypto.35.dylib
+                execute ln -s libcrypto.35.dylib libcrypto.dylib
+            execute popd
             build 'python' "Python-$PYTHON_BUILD_VERSION" ${PYTHON_BUILD_FOLDER}
             ;;
         *)
```

Have tried the same hack when building on macOS 10.15 too, but not entirely successful. Building is fine, stdlib `ssl` module seems fine, but loading `cryptography` for testing errors out with

```
Abort trap: 6 
````

Therefore, I will keep using `macos-1013` as default macOS builder for now.

**Drive-by fixes**:
  * Updated deps for `rhel-8` (also documented them in spreadsheet and csv files)
  * Patch Python to build on OpenBSD 6.6.
  * Handle files larger than 2GB on AIX.

reviewers: @adiroiban 

Please review changes.

Run the automated test, including `macos-1013`, `rhel-8`, and `obsd`.

No need to update runtime version in other repos, as I have generated required packages manually for RHEL 8 and OpenBSD 6.6 with the current revision in master.


dumol:
**needs-review**

dumol:
> I think that we can try using this python-package build in the chevah/server RHEL8 branch.

Not really needed, as the only changes related here to RHEL 8 are for deps checking. There is already a "backported" RHEL Python package for the current revision in master.

PR #143, opened 2020-07-03
Generic Linux ARM64 package.

dumol:
Scope
=====

Using the Pine64 board again these days, to keep my flat cool. 

But with a Debian-based Armbian this time.

Also as a way to test backup and recovery from scratch.

Changes
=======

Used the SD card with the Ubuntu 16.04 image to build a generic Linux `arm64` package by:

  1) Setting Ubuntu 16.04 as unsupported in `paver.sh` to trigger the generic Linux build.
  2) Adapting the generic `x64` Linux build for `arm64`.

**Drive-by fixes**:
   * Updated `cryptography` to latest version: 2.9.2.
   * Updated `setuptools` to latest version for Python 2.7x: 44.1.1.
   * Minor Alpine adjustments for 3.12 (previous PR was for 3.11).
   * Minor compilation fix for Debian 10 on `arm64`.
   * Travis CI integration for ARM64 builds.  
   * Avoid dangling `bin/bz*` symlinks in the install folder.

How to try and test the changes
===============================

reviewers: @adiroiban 

Please review changes.

Run automated tests, esp. the new Travis CI build for ARM64.

PR #144, opened 2020-07-09
[#5442] Brink transition.

dumol:
Scope
=====

Transition to brink, as the other repos: server, compat, etc.

Changes
=======

Quick and dirty adaptation for `chevah_build`.

**Drive-by fixes**:
  * Updated PIP to latest version: 20.1.1.
  * Updated Cython to latest version: 0.29.21
  * Updated `get-pip.py` to same version as PIP, to please my OCD.
  * Run compat tests on `fbsd-11`, now that they were fixed in `compat` repo.
  * Updated `pavement.py`
  * Added GitHub Actions for a few OS'es as an experiment.

reviewers: @adiroiban 

Review changes.

Run the automated tests.


dumol:
Alpine deps test is fixed in https://github.com/chevah/python-package/pull/143.

**needs-review**

adiroiban:
There is another cleanup that should be done.

Right now, I need the following workaround as otherwise the build fails... due to bzgrep and other broken symlinks

I think that python-package should remove these files before creating the distributable.

```
    # Remove the build helpers.
    if pave.fs.exists([target, 'lib', 'config']):
        pave.fs.deleteFolder([target, 'lib', 'config'])
```

adiroiban:
I would like to use this PR to also experiment with GitHub actions and see how we can implement a publish step from GitHub to our site.

One option is to have SFTPPlus running on weeps and have a special account which could upload files with HTTP (maybe on a non-standard port) and that account to be configured to only upload files)

I can setup GitHub action to upload the distributable after a successful build.

adiroiban:
I want to migrate from Buildbot to GitHub Actions as GitHub Actions has more developers and some nice features like for example cancelling an existing test if a newer version was pushed.

chevah_build
```diff
@@ -139,15 +139,20 @@ BUILD_FOLDER='build'
...
+CHEVAH_VARS=$(bash ./brink.sh output_chevah_vars 2>/dev/null)
 if [ "$?" -ne 0 ]; then
+    (>&2 echo 'Could not get CHEVAH_* variables from ./brink.sh!')
     exit 100
 fi
 
-PYTHON_VERSION=`cut -d' ' -f 2 DEFAULT_VALUES`
-OS=`cut -d' ' -f 3 DEFAULT_VALUES`
-ARCH=`cut -d' ' -f 4 DEFAULT_VALUES`
+IFS=' ' read -a chevah_vars_array <<< "$CHEVAH_VARS"
```
adiroiban: I think that we can do better here.
Store all the variables in a separate file and `source` that file inside the script.

This whole handling looks complicated.

dumol: Wait, we just got rid of `DEFAULT_VALUES` and now we are back to using a file? :-]

chevah_build
```diff
@@ -443,7 +443,7 @@ case $OS in
         EXTRA_LIBRARIES=""
         # A patched PyCrypto is packed and uploaded to our PyPI server.
         PIP_LIBRARIES="$PIP_LIBRARIES \
-            pycrypto==${PYCRYPTO_VERSION}chevah1 \
```
dumol, line 458: Sorry, but why would you not want our patched PyCrypto? It has quite a few security patches…

adiroiban: This was just to align with the other libraries.
As on Linux we were not using the patched version.

This needs to be fixed via the `${PYCRYPTO_VERSION}` which should include our patched version.

https://github.com/chevah/python-package/pull/144/files#diff-02f039358640626a68241789f3e4fc2dL32

So instead of using 

```sh
PYCRYPTO_VERSION=1.2.3
pip install pycrypto==${PYCRYPTO_VERSION}chevah1
```

we should have something like


```sh
PYCRYPTO_VERSION=1.2.3.chevah1
pip install pycrypto==${PYCRYPTO_VERSION}
```

Does it make sense?

dumol: We are using the patched version on Linux and other OS'es as well. It's just that not through PIP, as on Windows.

adiroiban: True.  I think that we can install pycryo via pip for all OS-es.

We should remove EXTRA_LIBRARIES as it introduced a long time ago when building with pip was complicated and should be kept only for the packages which don't work with pip.

-------

Also, I think that we should remove `PIP_LIBRARIES_OPENSSL_098` and just use the embedded openssl for those OSes.

There are so many exceptions and branches for how pachages are selected and I find it hard to read the code from here.

dumol:
> I am not sure why buildbot tests are no longer trigerred.

They are not automatically triggered in this repo for every commit in every branch, only for `master` branch. This is because they take quite a lot, especially on the slower slaves. Imagine the havoc after triggering builds on all slaves marked for review for every commit you've tried with GitHub actions… :-]

I have triggered a `test_review` now.


> The builds on GitHub action works for both MacOS and Windows 2019.
> 
> On MacOS they fail as the OS has gettext installed. I think that our buildscript should be updated to not compile gettext, even if gettext library is present. If I remove gettext, then all other tools like wget will fail.

I've tried removing Homebrew altogether before starting building, but it doesn't seem to improve on this. Worse thing, I can't debug it because I can't login to see what linked to `/usr/local/opt/gettext/lib/libintl.8.dylib`.
 
> On Windows, the tests are failing at compat step.
> 
> I hope that you can find out and fix buildbot tests.

Not sure here, do you mean fixing the Windows compat tests on GitHub? On our Buildbot they run just fine.

> I think that this can be merged as it is.
> 
> Once we have SFTPPlus installed on the download server / binary.chevah.com I can update GitHub action to automatically copy the build files for testing.

You mean automatically copy the packages? But what about the platforms that GitHub doesn't have? What about generic Linux which is build on SLES 11? Even with self-hosted runners, the choice of platforms is quite limited: https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners

adiroiban:
> They are not automatically triggered in this repo for every commit in every branch, only for `master` branch.

Thanks. Makes sense.

----------




> I've tried removing Homebrew altogether before starting building, but it doesn't seem to improve on this.

Yes... debugging is complicated.
The plan is to add support for tmux debugging which will be enabled when a commit contains the `[debug]` word in the commit message.

Debugging is easier when we run on our hosted VMs.


This was just an experiment to discover the issues we have with GitHub actions.

This is also a plan B in case we  have serious problems with our datacenter.

--------

> Not sure here, do you mean fixing the Windows compat tests on GitHub? On our Buildbot they run just fine.

I need to see why they are green on buildbot... and why they pass here.. but this is a minor issue.

----------

> You mean automatically copy the packages? But what about the platforms that GitHub doesn't have?

We should update our chevah_build script to be able to compile a generic/standalone/static-linked python even when compiled on Ubuntu.
This can be trigerred via an env variable.

I don't plan to fully replace our own hosted slave VM.
But I do plan to replace buildmaster and use GitHub Action runners instead of buildslaves.

Another option is to use Docker.. GithHub Action cloud environment support running things inside docker.
In theory we can use that to build for Alpine.  

----------

I tried to load the ENV vars from the fixed `build` dir used by chevah_build

dumol:
> We should update our chevah_build script to be able to compile a generic/standalone/static-linked python even when compiled on Ubuntu.
> This can be trigerred via an env variable.

Not sure this is a good idea… We generate the generic Linux x64 build on SLES 11 because it has to run on everything from SLES 11 and RHEL 6 to latest Arch.

chevah_build
```diff
@@ -138,8 +138,9 @@ BUILD_FOLDER='build'
 
 # Import shared code.
 . ./functions.sh
-# Import build environmet variables.
-. ./build-python-package/BUILD_ENV_VARS
+
+# Import build environment variables.
+. ./${BUILD_FOLDER}/BUILD_ENV_VARS
```
dumol, line 143: In `./brink.conf` this is always `build`, so theoretically-speaking this might go out of sync.

But why all this toil to use a sub-dir? Let's just save to a file in the current dir and be done with it, `DEFAULT_VALUES` worked for years… :-]

adiroiban: I didn't wanted to "pollute" the root dir and as a best practice have everying inside the build dir.

adiroiban:
> Not sure this is a good idea… We generate the generic Linux x64 build on SLES 11 because it has to run on everything from SLES 11 and RHEL 6 to latest Arch.

Well.. this is a plan B idea... but in theory the generic linux should work on any modern linux, including Ubuntu 20.04.
There is also Ubuntu 16.04 which is a bit older

.github/workflows/main.yaml
```diff
@@ -93,10 +93,14 @@ jobs:
 
     # Make the OS as close as possible to a standard one.
     # But make sure we have the required dev libraries.
-    - name: Prepare OS
+    - name: Remove Homebrew
```
adiroiban, line 96: My goal was to have all the OS preparate task in a step... there should a balance between putting all tasks in a step, or creating a separate step for each executed command.

I prefer to have multiple steps, each step grouping multiple command that have the same purpose.

dumol: ACK.

chevah_build
dumol: Agreed, this is already spelled out in the ticket for a Python 3 package. But here I don't want to do too many things at the same time. For now, I have reversed your changes for Windows, Linux, and AIX 5.3.

Besides, I want to do those cleanups in dedicated PRs, so that they are easy to find in history. Maybe before forking for Python 3. Although I'm not so sure a forking is needed, we could just get rid of all the external sources and then advise using shallow clones.

PR #145, opened 2020-11-04
[#5508] Cryptography 3.2 and other updates.

dumol:
Scope
=====

Fix CVE-2020-25659.

Changes
=======

Updated `cryptography` to 3.2.1.

Other updates to latest versions:
  * OpenSSL 1.1.1g -> 1.1.1h
  * pip 20.1.1 -> 20.2.4
  * cffi 1.14.0 -> 1.14.3
  * psutil 5.7.0 -> 5.7.3
  * pywin32 227 -> 228.

**Drive-by fixes**:
  * Allow latest cryptography with OpenSSL 1.0.2 on specific platforms: RHEL 7, AMZN 2, AIX,  Solaris 11.4.
  * Suppress specific cryptography warnings in regards to Python 2 and OpenSSL 1.0.2.
  * Build cryptography on the generic Linux platform (currently CentOS 5.11). There's no need for the upstream wheel any more, since switching for this platform from SLES 11.4.
  * OpenBSD 6.8 and FreeBSD 12 support.
  * Brink updates from compat/server repos.

reviewers: @adiroiban 

Please review changes excluding source updates, e.g.:
```
git diff master brink.* chevah_build src/*/chevahbs python-modules/chevah-python-test/
```


dumol:
This now depends on https://github.com/chevah/compat/pull/103 for the platforms with OpenSSL 1.0.2 (Amazon 2, RHEL 7, Solaris 11.4).

chevah_build
```diff
@@ -998,7 +998,7 @@ command_test_compat() {
     # the local version.
     echo '##### Running chevah.compat tests... #####'
     execute rm -rf compat
-    execute git clone https://github.com/chevah/compat.git --depth=1 -b master
+    execute git clone https://github.com/chevah/compat.git --depth=1 -b 5509-cryptography32-openssl102
```
adiroiban, line 1003: the code is in master now :)

adiroiban:
I have enabled the latest python as part of this PR https://github.com/chevah/server/pull/1735


It doesn't look that bad on my laptop but it fails with missing space on the slaves:

https://chevah.com/buildbot/builders/server-ubuntu-2004/builds/957

adiroiban:
I think the cryptography env issue should be fixed by patching the `site.py` file and set the env there.
You can set the env in all python packages, or maybe just into amz2 and rhel7.

There is another thing that needs to be "patched"

```
$ ./build-server/bin/python 
Python 2.7.18 (f321735d, Nov  9 2020, 18:58:17) 
[GCC 9.3.0] on linux2
>>> import cryptography
__main__:1: CryptographyDeprecationWarning: Python 2 is no longer supported by the Python core team. Support for it is now deprecated in cryptography, and will be removed in a future release.
>>> 

$ ./build-server/bin/python 
Python 2.7.18 (f321735d, Nov  9 2020, 18:58:17) 
[GCC 9.3.0] on linux2
>>> import OpenSSL
/home/adi/chevah/server/build-server/lib/python2.7/site-packages/OpenSSL/crypto.py:12: CryptographyDeprecationWarning: Python 2 is no longer supported by the Python core team. Support for it is now deprecated in cryptography, and will be removed in a future release.
  from cryptography import x509

```


The path is to ignore the warnings

```
$ ./build-server/bin/python 
Python 2.7.18 (f321735d, Nov  9 2020, 18:58:17) 
[GCC 9.3.0] on linux2
>>> import warnings
>>> warnings.filterwarnings('ignore', category=UserWarning, module='OpenSSL.crypto')
>>> import OpenSSL
>>>
```

In site.py we should have something like this beside the os.environ change.

```
import warninigs
warnings.filterwarnings(
    'ignore',
    category=UserWarning,
    module='OpenSSL.crypto',
    )
```

adiroiban:
With these warnings error, I think that we should update the way chevah_build test to put all stderr into a file and fail if the file is non-empty at the end of run

adiroiban:
Thanks.

it looks good, but we need to fix the warnings.

A server run
https://chevah.com/buildbot/builders/server-amzn-2/builds/1306/steps/test/logs/stdio

I think that this should be fixed in python site.py and not in chevah/compat

The required fixed for site.py

* Ignore cryptography py2 warning - all OSes
* Set cryptography env for older OpenSSL - can be set for all OSes if is easier.

neeed-changes

chevah_build
dumol: Thanks, I was waiting to see all OS'es run `compat` tests fine. Happy to have found the little fix required to fix them on AIX.

dumol:
> I think the cryptography env issue should be fixed by patching the `site.py` file and set the env there.
> You can set the env in all python packages, or maybe just into amz2 and rhel7.

So it's not enough is set in compat? We have to set it here too? (Then maybe we can remove it from compat after the latest python packages are used there?)

In compat the required env was exported unconditionally. The OS'es with OpenSSL 1.0.2 are:
  * RHEL 7
  * AMZN 2
  * AIX
  * Solaris 11.4
 
> There is another thing that needs to be "patched"

[…]

> 
> In site.py we should have something like this beside the os.environ change.
> 
> ```
> import warninigs
> warnings.filterwarnings(
>     'ignore',
>     category=UserWarning,
>     module='OpenSSL.crypto',
>     )
> ```

Fine by me, but wouldn't this be kind of dangerous? I gather this suppresses all warnings from `cryptography`…

dumol:
> With these warnings error, I think that we should update the way chevah_build test to put all stderr into a file and fail if the file is non-empty at the end of run

A good idea in theory, but in practice we have errors we can't really get rid of. E.g.:

 * loading our heavily-patched `readline` returns `bind: Invalid command 'enable-meta-key'.` 
 * `pypi` keeps outputting: `DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. pip 21.0 will drop support for Python 2.7 in January 2021. More details about Python 2 support in pip can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support pip 21.0 will remove support for this functionality.`
* even `scandir`'s test output always goes to stderr.

It's quite messy.

dumol:
Implemented the above as discussed in private.

At least one more warning remains, on Solaris 10, don't know how to deal with it:
```
/srv/buildslave/runtime/build-solaris10-x64/slave/python-package-sol-10/build/build/python2.7-sol10-x86/lib/python2.7/site-packages/subprocess32.py:149: RuntimeWarning: The _posixsubprocess module is not being used. Child process reliability may suffer if your program uses threads.
  "program uses threads.", RuntimeWarning)
```

For the moment, haven't bothered with the OS'es not included in `test_review`, e.g. AIX and HP-UX.

**needs-review**

python-modules/chevah-python-test/test_python_binary_dist.py
```diff
@@ -648,6 +648,21 @@ def main():
     else:
         print 'stdlib ssl %s' % (OPENSSL_VERSION,)
 
+    try:
+        # Import OpenSSL first to suppress cryptography warnings effectively.
+        import OpenSSL
+        from OpenSSL import SSL, crypto, rand, __version__ as pyopenssl_version
```
adiroiban, line 654:
```suggestion
        from OpenSSL import crypto, SSL, rand, __version__ as pyopenssl_version
```

just a suggestion... maybe this is enough

src/python/cryptography_silent_warnings.patch
```diff
@@ -0,0 +1,21 @@
...
++warnings.filterwarnings(
++    'ignore',
++    category=UserWarning,
++    module='OpenSSL.crypto',
++    )
++warnings.filterwarnings(
++    'ignore',
++    category=UserWarning,
++    module='cryptography.hazmat.bindings.openssl.binding',
++    )
```
adiroiban: I found another way to ignore these warnings....and I think that this is the better approach

```
warnings.filterwarnings(
    'ignore',
    category=UserWarning,
    message='EXACT MESSAGE TO BE IGNORED',
    )
```

We should ignore both py 2.7 and openssl 1.0.2

This is better than filtering based on module, since the cryptography error is raised based on whichever other module is importing it.

Thanks

dumol: Thanks, that would be much better. Going to give it a try…

python-modules/chevah-python-test/test_python_binary_dist.py
dumol: Good catch! But this not needed any more when filtering exact messaging, so I'm going to revert the change.

adiroiban:
Looks good. Great work.

Please push it into prod.

changes-approved

PR #148, opened 2021-06-30
[#5577] Smaller packages and other cleanups.

dumol:
Scope
=====

Backport the package cleanups from pythia to halve the size of the resulting packages.

Including removing pycryptodomex and cython. And pycrypto.

Also, backport the downloading code from chevah/pythia to always use curl to download from GitHub's servers (from published releases).

In addition, add retrying, as exemplified at ​https://stackoverflow.com/a/42873372.

Changes
=======

Removed GMP, `gmpy2`, `pycrypto`, `pycryptodomex`, `cython`.

Backported the package clean-ups from Pythia. Adapted them for Python 2.7, especially on Windows, where there's lots of cruft compared to the minimal Python 3.8 packages.

Backported all `brink.sh` improvements from Pythia. GitHub releases is the default source for downloading Python packages now. `curl` is the only supported downloader.

Added retrying options for `curl` for both checking and downloading online packages. Only options that work on CentOS 5 are used.

`PYTHON_PACKAGE_VERSION` file now also contains OS and ARCH, just like a package name. While at it, only use 7 characters for the Git revision, like GitHub and Pythia. Also use the Pythia-way of determining it, for consistency across GHA workflows.

**Drive-by fixes**:
  * Alpine 3.14 support.
  * Minor Python updates from ActiveState, fixes CVE-2021-23336. Have stopped short of picking their 2.7.18.x versioning because on Windows we are still using upstream packages with (almost) all their security fixes applied as hot patches. (The exception is CVE-2021-3177, which requires recompiling: https://github.com/ActiveState/cpython/commit/8bf77ab57ea944814d69bf7f983561cc26bcf96e)
  * SQLite updated to 3.36.0.
  * Visual C for Python 2.7 is now downloaded from https://binary.chevah.com and installed separately because there's no longer a Chocolatey package as Microsoft has removed the download from their site.
  * Alpine / RHEL 8 / Generic Linux packages are now built through Docker.
  * ARM64 is still built through our Pine64 board, but using a GHA runner associated to this repo. I start it manually only when I need it, so there's a low chance of it being exploited. Even then, we should be notified of the fork and I can reflash the SD card.
  * AIX 7.1 package is built manually on Eliseni on the WPAR set up with the IBM compiler.

reviewers: @adiroiban 

To review relevant changes: `git diff master .github/ brink.* chevah_build functions.sh publish_dist.sh python-modules/chevah-python-test/ src/*/README src/*/chevahbs`

https://bin.chevah.com:20443/testing/2.7.18.71afee7/  was tried for chevah/server in https://github.com/chevah/server/pull/1822.


dumol:
**needs-review**

adiroiban (approved):
Great work.
So far the chevah/server PR looks ok so I think that we can merge it.

If we found any issues is best to fix them in a separate PR

PR #149, opened 2021-07-16
Master branch workflow fixes.

dumol:
Scope
=====

Fix cloning sources independently for both master and PR branches.

Changes
=======

Adapted the relevant code from chevah/server.

**Drive-by fixes**:
  * Clone sources independently on all platforms, for consistency.
  * Make sure `stderr` output is always shown for our Python tests.

reviewers: @adiroiban

### DIFF
--- a/.github/workflows/bare.yaml
+++ b/.github/workflows/bare.yaml
@@ -44,6 +44,7 @@ jobs:
     # GHA's checkout action seems to be misbehaving on ARM64 and consequently
     # our build script can't get current revision from cloned sources.
     - name: Clone sources independently
+	  if: matrix.runs-on == 'ARM64'
       run: |
         git init $CHEVAH_REPO
         cd $CHEVAH_REPO
@@ -56,7 +57,7 @@ jobs:
         git fetch --no-tags --prune origin
         # Prepare the code.
         git clean -f
-        git reset --hard ${{ github.event.after }}
+        git reset --hard ${{ github.event.after }} || true
 
     - name: Prepare OS
       if: startsWith(matrix.runs-on, 'ubuntu')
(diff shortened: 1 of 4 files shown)


.github/workflows/bare.yaml
```diff
@@ -38,16 +39,25 @@ jobs:
       uses: actions/checkout@v2
       # Required to get the correct git rev to use in Python's version.
       with:
-        fetch-depth: 2
+        ref: ${{ github.event.after }}
```
adiroiban, line 42: Minor comment
can we remove the usage of `actions/checkout@v2` and use `Clone sources independently` everywhere?

.github/workflows/bare.yaml
```diff
@@ -38,16 +39,25 @@ jobs:
...
 
     # GHA's checkout action seems to be misbehaving on ARM64 and consequently
     # our build script can't get current revision from cloned sources.
-    # With the repo cloned in place beforehand, just checkout current branch.
-    - name: Checkout branch for already cloned sources
+    - name: Clone sources independently
```
adiroiban, line 46: minor comment 
I think that if this is only using environment variable, it can be extracted into a separate  "composite-step" directory+file (like https://github.com/chevah/server/blob/master/.github/actions/prepare-run/action.yaml)

and reuse in all other places... 

to have a single way of getting the code.

adiroiban (approved):
all good. thanks!

.github/workflows/bare.yaml
dumol: In the end, I did go for cloning sources independently everywhere, including macOS and Windows. For a lower mental load associated with all this… Thanks!

PR #150, opened 2021-07-19
Fixed publishing on Windows.

dumol:
Scope
=====

Publishing on Windows was broken by a "cosmetic" change.

Changes
=======

Switch back to the old sequence of commands for publishing on Windows.

**Drive-by fix**:
  * Actually fail when publishing a package fails. At the expense of leaving private keys behind on disk.

reviewers: **WIP**

### DIFF
--- a/.github/workflows/bare.yaml
+++ b/.github/workflows/bare.yaml
@@ -72,7 +72,6 @@ jobs:
         ./chevah_build compat
 
     # Upload using a (per-OS selected) sftp command, then show final links.
-    # Remove key in same step to have it deleted even if publishing fails.
     - name: Upload testing package
       run: |
         mkdir -pv ~/.ssh/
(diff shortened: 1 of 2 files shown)


.github/workflows/bare.yaml
```diff
@@ -38,8 +38,8 @@ jobs:
       if: startsWith(matrix.runs-on, 'ubuntu')
       run: sudo apt-get install -y libncurses5-dev
 
-    # GHA's checkout action seems to be misbehaving on ARM64 and consequently
-    # our build script can't get current revision from cloned sources.
+    # Git's version on our ARM64 runner is too old for actions/checkout to
+    # create a local repo. Beware this fails on opening a new PR.
     - name: Clone sources independently
       run: |
         git init $CHEVAH_REPO
```
adiroiban, line 45: I think that we should add a final "git log -1 --format='%H'" to show the current HEAD .
This is what actions/checkout does and I forgot to add it.

dumol: But this would just show something, how it would fix anything?

adiroiban: It will not fix...is just something that I found that actions/checkout does and the custom script is not doing.

It is there only to help with debugging.

And I think that an `ls -al .` can also help  with debugging.

I will experiment with it on chevah/server and will let you know how it goes.

dumol: Thanks! Is this the reason you have not approved this? Or what is preventing it?

adiroiban (approved):
All good. Thanks!

This can be merged.

the review was delay due to lack of time. Sorry for the delay.

.github/workflows/bare.yaml
dumol: I've added your suggested command. Thanks!

PR #158, opened 2021-11-02
Use patched pip to solve latest security issue.

### LINKED ISSUE #159: Cleanup the distributed package
For python-package I think that the final "packaging script" should be updated to do the following

* remove /man directory
* move /include into lib/includ
* move /pkgconifg into the existing /lib/pkgconfig

dumol:
Scope
=====

Safety tests fail in master with:
```
> pip, installed 20.3.4, affected <21.1, id 42218
Pip version 21.1 stops splitting on unicode separators in git references, which could be maliciously used to install a different revision on the repository. 
https://github.com/pypa/pip/issues/9827
```

Changes
=======

Backported the fix from https://github.com/pypa/pip/pull/9827/files for pip 20.3.4 and built a wheel versioned as `20.3.4chevah1`.

I've also tried version `20.3.5chevah1`, but it didn't help in getting rid of the upgrade warning:
```
WARNING: You are using pip version 20.3.5chevah1; however, version 20.3.4 is available.
```

**Drive-by fixes**:
  * Updated OpenSSL 1.1.1 sources to version 1.1.1m.
  * Updated SQLite, pyOpenSSL, `cffi`, `psutil`, `pycparser`.
  * Fixes #159.

reviewers: @adiroiban 

Evaluate the fix from https://github.com/pypa/pip/pull/9827/files.

Run the automated tests.


adiroiban:
> WARNING: You are using pip version 20.3.5chevah1; however, version 20.3.4 is available.

The version should be something like 20.3.4+chevah1 ...but version comparison is undefined for custom versions.

But I think that we should ignore the pip warning.
as long a safety is ok, we should also be ok :)

python package versioning is defined in https://www.python.org/dev/peps/pep-0440/

dumol:
Thanks, I was looking at pip's history and tried to follow the same convention. Have only found things like 20.2b1 and 10.0.0b1.

chevah_build
```diff
@@ -30,7 +30,8 @@ PSUTIL_VERSION="5.8.0"
 SUBPROCESS32_VERSION="3.5.4"
 
 # Versions no longer upgradable because of Python 2 deprecation.
-PIP_VERSION="20.3.4"
+# Backported fix for https://github.com/pypa/pip/issues/9827.
+PIP_VERSION="20.3.4chevah1"
```
adiroiban, line 34: Where is the code for this release ?

I don't see it at https://github.com/chevah/pip/branches

adiroiban (changes requested):
Thanks for the update.

I think this PR is missing the link to the actual changes in pip.

chevah_build
dumol: It's at https://github.com/pypa/pip/pull/9827/files, as linked in the PR description. There is no code from me.

Do you want me to re-fork https://github.com/chevah/pip? The current fork is 9 years old.

dumol: Did a "Fetch upstream" to get our fork up to date. Unfortunately, there's no branch for Python 2.7 compatibility.

What to do? Hard-reset our fork to something that was still compatible and branch out from there? 

At https://pypi.org/project/pip/#history, I see version 20.3.4 was released on January 23, 2021. Corresponding commit would be https://github.com/chevah/pip/commit/a27d06eecf12f16fab7ae78b0e0889745c7a4436.

dumol: OK, I've got something at https://github.com/chevah/pip/tree/20.3.4chevah. Good enough? Anything else? Thanks.

adiroiban (approved):
Thanks. 

I saw the diff, but I was looking for the source for the exact `pip-20.3.4chevah1`  package.

The one from :)
https://bin.chevah.com:20443/pypi/simple/pip/pip-20.3.4chevah1-py2.py3-none-any.whl

I think that it helps to put the link to our GitHub fork.

and in setup.py use `version=get_version("src/pip/__init__.py") + 'chevah1'`,
https://github.com/chevah/pip/blob/20.3.4chevah/setup.py#L31

All good. Thanks!

dumol:
I've tried to double-check all is fine, so I have regenerated the wheel from the `20.3.4chevah` branch of chevah/pip, to which I've added a few more commits today.

For the record, this is what I've used on Ubuntu 20.04:
```
python3 setup.py bdist_wheel --universal
```

However, this is as not as good as the previous one generated from patching the 20.3.4 release... I get:

```
Collecting pip==20.3.4chevah1
  Downloading https://bin.chevah.com:20443/pypi/simple/pip/pip-20.3.4chevah1-py2.py3-none-any.whl (1.5 MB)
ERROR: Package 'pip' requires a different Python: 2.7.18 not in '>=3.6'
PWD : /home/runner/work/python-package/python-package/python-package
Fail: /home/runner/work/python-package/python-package/python-package/build/python2.7-ubuntu1804-x64/bin/python python-modules/get-pip.py --index-url=https://bin.chevah.com:20443/pypi/simple --trusted-host=bin.chevah.com --no-cache-dir --no-warn-script-location pip==20.3.4chevah1 --no-setuptools setuptools==44.1.1
Error: Process completed with exit code 1.
```

For now, I have reverted to a wheel generated from the patched released tarball, for which I've used version `20.3.4chevah0`. Let me know what could be wrong, I've used exactly the same changes and the same command as above. I've double-checked and regenerated it, to be sure. Thanks!

**needs-review**

adiroiban:
The current pip package information defined as 3.6 is the minimum version

via https://github.com/chevah/pip/blob/20.3.4chevah/setup.py#L85

-------------

It was introduced here https://github.com/chevah/pip/commits/7c775c86df34bffd3b56df58847ca87f8bd5778f

i see that 20.3.4 is the last version with support for 2.7 - https://github.com/pypa/pip/tree/20.3.4

So the patch should be applied on the code from that tag 

Make sense?

dumol:
Thanks! Our fork didn't have that tag. I've tried yesterday to use what looked to me like the relevant commit, but it definitely wasn't the right stuff.

I have re-forked our repo from command line through changing URL remote, after resetting to that tag. Branched out from there, re-applied the changes, and regenerated the `chevah1` wheel. This time it's all good…

dumol:
As this has stalled for quite a while, should I also update embedded OpenSSL libs to latest 1.1.1 version, as a drive-by fix?

Bugfixes only this time for 1.1.1m, but still… More at https://www.openssl.org/news/cl111.txt

adiroiban:
All good. Thanks and sorry for the delay.

yes. I think that updating openssl would help.

adiroiban:
needs-changes

adiroiban:
Thanks. I will wait for the files to hit the testing server and I will give them a try.

adiroiban (approved):
I think that we should merge it.

adiroiban:
Thanks.
I think that we should merge it.
If we found any issue while running chevah/server tests we can fix it in a separate PR...as this PR is quite big.

